Issue #5, opened 2018-09-13
Add support for tasks which run different test filters


runningcode:
Here are my thoughts on the API:
~~I was thinking of adding a map to the extension like so:~~
```
additionalTestFilters = mapOf(
  "pr" to listOf(listOf("class .........", "package......."))
  "regression" to listOf("class .........", "package......."))
)
```

~~this would generate two additional tasks~~
~~`runFlankPr` and `runFlankRegression` in addition to the `runFlank` task which would run with the default `testFilters`.~~

Let's use customizable tasks instead.

winterDroid:
Ah yeah I see what you mean. Honestly I think that is not really necessary since that is what customizable tasks are made for, right?

PR #12, opened 2018-09-26
Create a new RunFlankTask class.

runningcode:
This will allow us to pass arguments and configure the task in the
future. Reference #5.

This task is composed of other tasks which download flank, create a yaml
file and run the flank command line task.

### DIFF
--- a/buildSrc/src/main/java/com/osacky/flank/gradle/RunFlankTask.kt
+++ b/buildSrc/src/main/java/com/osacky/flank/gradle/RunFlankTask.kt
@@ -1,0 +1,13 @@
+package com.osacky.flank.gradle
+
+import org.gradle.api.DefaultTask
+import javax.inject.Inject
+
+open class RunFlankTask @Inject constructor(private val extension: FlankGradleExtension) : DefaultTask() {
+
+  init {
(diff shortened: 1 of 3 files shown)


buildSrc/src/main/java/com/osacky/flank/gradle/RunFlankTask.kt
```diff
@@ -0,0 +1,11 @@
+package com.osacky.flank.gradle
+
+import org.gradle.api.DefaultTask
+import javax.inject.Inject
+
+open class RunFlankTask @Inject constructor(private val extension: FlankGradleExtension) : DefaultTask() {
```
runningcode: @winterDroid In this pull request, I moved the run flank task to a separate class so we can add parameters to it in the future.
This class `dependsOn` the actual tasks which will download and run flank.
We'll have to figure out how to pass parameters from this task to its dependencies, but I think this is the right starting point. What do you think @winterDroid ?

winterDroid: Yeah that's probably a good starting point 👍

PR #15, opened 2018-10-17
Add support for tasks which run different test filters.

### LINKED ISSUE #5: Add support for tasks which run different test filters

runningcode:
Use NamedDomainObjectContainer to allow the tasks to be named.

Fixes #5

### DIFF
--- a/buildSrc/src/main/java/com/osacky/flank/gradle/FladleConfig.kt
+++ b/buildSrc/src/main/java/com/osacky/flank/gradle/FladleConfig.kt
@@ -1,0 +1,17 @@
+package com.osacky.flank.gradle
+
+interface FladleConfig {
+  var flankVersion: String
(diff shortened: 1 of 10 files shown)


runningcode:
What do you think @winterDroid ?

Also thanks to @vanniktech for the idea!

winterDroid:
LGTM, but I'd rename `targets` to something like `config` and would add some tests that the proper tasks are created.

PR #36, opened 2019-02-27
Use Flank as a maven artifact.

### LINKED ISSUE #3: Download flank to build cache folder
So as not to redownload it every time

runningcode:
This removes the need to redownload flank every time.
This also removes the snapshot dependencies and the Download task
dependencies.

Fixes #3

### DIFF
--- a/buildSrc/src/main/java/com/osacky/flank/gradle/FlankGradlePlugin.kt
+++ b/buildSrc/src/main/java/com/osacky/flank/gradle/FlankGradlePlugin.kt
@@ -2,19 +2,32 @@ package com.osacky.flank.gradle
 
 import com.android.build.gradle.AppExtension
 import com.android.builder.model.TestOptions
-import de.undercouch.gradle.tasks.download.Download
 import org.gradle.api.GradleException
 import org.gradle.api.Plugin
 import org.gradle.api.Project
+import org.gradle.api.artifacts.Configuration
 import org.gradle.api.tasks.Exec
 import org.gradle.api.tasks.TaskContainer
+import org.gradle.kotlin.dsl.repositories
 import org.gradle.util.GradleVersion
 
 class FlankGradlePlugin : Plugin<Project> {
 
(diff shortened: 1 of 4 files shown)


buildSrc/src/main/java/com/osacky/flank/gradle/FlankGradlePlugin.kt
```diff
@@ -69,17 +70,17 @@ class FlankGradlePlugin : Plugin<Project> {
       description = "Finds problems with the current configuration."
       group = TASK_GROUP
       workingDir("${project.fladleDir}/")
-      commandLine("java", "-jar", "flank.jar", "firebase", "test", "android", "doctor")
-      dependsOn(named("downloadFlank"), writeConfigProps)
+      commandLine("java", "-jar", project.fladleConfig.singleFile.absolutePath, "firebase", "test", "android", "doctor")
```
winterDroid: Can't you instead set the class path to the configuration and then use args instead? Then you should not need to use the path to the jar directly

buildSrc/src/main/java/com/osacky/flank/gradle/FlankGradlePlugin.kt
```diff
@@ -2,18 +2,30 @@ package com.osacky.flank.gradle
...
+    // Add Flank maven repo.
+    target.repositories {
+      maven { url = target.uri("https://dl.bintray.com/flank/maven") }
+    }
+
+    // Create Configuration to store flank dependency
+    target.configurations.create(FLADLE_CONFIG)
+
+    // Add Flank dependency to Fladle Configuration
+    target.dependencies.add(FLADLE_CONFIG, "flank:flank:4.4.0")
```
winterDroid: Shouldn't that version be configurable?

runningcode: zomg :)

PR #53, opened 2019-05-17
52: `shardTime` config property not written to flank.yml

nnoel-grubhub:
While shardTime is present in the gradle plugin's configuration properties, it is not supported in FlankGradleExtension or YamlWriter. It will not be written to flank.yml properly as shard-time
This prevents the ability to use shardTime to dynamically set the shardCount.


runningcode:
Thank you very much for the contribution!

PR #68, opened 2019-08-29
Add support for additonal test apks.

### LINKED ISSUE #67: Add support `additional-app-test-apks`
See https://github.com/TestArmada/flank/pull/542 for more information.

runningcode:
Fixes #67

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.7.0
+ * Add support for additional test apks.
+ * Bump flank version to 7.0.2
+
 ## 0.6.6
 * Bump flank version to 7.0.0
 * Publish to mavenCentral()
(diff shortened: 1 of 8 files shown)


charlesdurham:
I think the common use case is a single app apk and then multiple test apks. Flank supports this by only requiring a single app apk and using that as the fallback if its omitted w/ later test apks. Might be nice to carry that functionality over here.

runningcode:
@charlesdurham hey, that's a great idea. can you help me structure the DSL to support this use case?
one way would be like
listOf("- app: app", "test: test.apk", "test: test2.apk"), its a bit more verbose than the current, but it might work better. what do you think?

charlesdurham:
Yeah it gets kind of messy and I can't think of anything great, maybe it's just better as-is.

The only thing I could think of was making the pair `Pair<String?, String>` instead. But then passing null is almost the same as repeating the path to the app apk 🤷‍♂.

runningcode:
I was also just thinking of having the whole thing be a string?

charlesdurham:
Up to you 😄 `List<Pair<String, List<String>>` could also work where the first string is the app apk and the list of strings are the test apks.

alosdev:
@runningcode will it be merged one day?

PR #72, opened 2019-09-20
Allow wildcards in directory names.

### LINKED ISSUE #71: Add ability to use wildcards in apk path
@winterDroid

runningcode:
This removes the check that the file exists. This also means that if the
file doesn't exist, flank may file with a weird cryptic error.

Fixes #71


winterDroid:
Amazing!

PR #75, opened 2019-10-18
Support Flank 8.x.x

francescocervone:
[Flank 8.0.0](https://github.com/TestArmada/flank/releases/tag/v8.0.0) added two breaking changes.
- `flaky-test-attempts` has been renamed to `num-flaky-test-attempts`
- `repeat-tests` has been renamed to `num-test-runs`

This PR adds support for Flank 8 maintaining support for older Flank's versions at the same time.

### DIFF
--- a/buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
+++ b/buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
@@ -1,5 +1,6 @@
 package com.osacky.flank.gradle
 
+import org.gradle.api.GradleException
 import org.gradle.internal.impldep.com.google.common.annotations.VisibleForTesting
 
 internal class YamlWriter {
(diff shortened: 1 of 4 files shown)


buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
```diff
@@ -107,4 +118,20 @@ internal class YamlWriter {
       }
     }
   }
+
+  private fun isFlankVersionAtLeast(flankVersion: String, major: Int): Boolean {
+    val matchResult = FLANK_VERSION_REGEX.matchEntire(flankVersion)
```
runningcode: This solution works, but I would prefer if we tried to use the more idiomatic `VersionNumber` api in gradle. Can you try to see if this same thing can be achieved with `VersionNumber.parse(flankVersion).getMajor() > major`?

francescocervone: Sure! Fixed in https://github.com/runningcode/fladle/pull/75/commits/db765b58b92d13b81aca7fe0365bc23daf9b9293

runningcode: Beautiful

runningcode:
Released in 0.7.0 https://github.com/runningcode/fladle/blob/master/CHANGELOG.md#070

Issue #31, opened 2019-02-26
Device not found


ronaldvanduren:
When trying to add a new device as an optional parameter, the device object is not found on the classpath com.osacky.flank.gradle.Device. I'm using the following fladle block: 

apply plugin: "com.osacky.fladle"

import com.osacky.flank.gradle.Device 

fladle {
    // Required parameters
    serviceAccountCredentials("${project.file("path-to-key-file.json")}")

    // Optional parameters
    useOrchestrator = true

    environmentVariables = ["annotation": "some-annotation"]

    directoriesToPull = [
            "/sdcard/screenshots"
    ]
    filesToDownload = [
            ".*/screenshots/.*"
    ]
    timeoutMin = 15
    recordVideo = true
    performanceMetrics = false
    devices = [
            new Device("Pixel2", 28, null, null),
    ]
    flankVersion("v4.3.1")
    debugApk("$buildDir/outputs/apk/debug/app-debug.apk")
    instrumentationApk("$buildDir/outputs/apk/androidTest/debug/app-debug-androidTest.apk" )
    autoGoogleLogin = true
    testShards = 15
}

I'm using Gradle 5.2.1 and Gradle Plugin 3.3.0.

runningcode:
Thanks for the report.
Can you copy a stacktrace here if possible?
What version of flank were you using?

runningcode:
Also can you add the code of where you added the fladle plugin to the classpath?

winterDroid:
@runningcode I have the same issue. But it seems to me just to be an IDE issue since it is still compiling just fine. Also happens to me with other Gradle plugins.

runningcode:
Thanks for filing. I'm going to close this as bug in Intellij/AS.

We can also alternatively propose a different API which wouldn't require a dependency on a class in the plugin to make it easier to use. Please feel free to make an issue with a suggestion!

PR #76, opened 2019-10-31
Map Device List

zlippard:
Updated `devices` to support a different API using a list of maps for setting up devices rather than requiring the `Device` class to be referenced in Gradle.

In reference to: https://github.com/runningcode/fladle/issues/31

Usage:

```groovy
fladle {
  ...
  devices = [
/*  No longer allowed  */
//    new Device("Nexus5", 28)
/*  Use this instead  */
    [ "model": "Nexus5", "version": "28" ]
  ]
}
```


runningcode:
Hey, thanks for the contribution. I really like this new api design. Can we deprecate the old one and only allow the new one to be used? Also, can we only allow one of the two to be set from the fladle extension? (Throw an exception if both are set)

zlippard:
@runningcode sure thing! I'll add the deprecation and force only one to be used in the extension.

zlippard:
@runningcode Do you think it would be better to simply _replace_ `devices` with the list map implementation instead of a list of `Device`s?  Otherwise, we're requiring the users of this to switch over to `mapDevices`.  I had gone back and forth on this, but I'm wondering if now it makes more sense to simply change `devices` to be the list map implementation.  It is a breaking change this way, but maybe this would be a major version change?  

Thoughts?

runningcode:
Yeah, you know what? Fladle is still a 0.x library so we reserve the right to make these changes. Let's go ahead and do that. I think the importing of the `Device` has been a really confusing API for some time now. Let's clean this up and have a single opinionated way of doing this. Let's have the new map field only.

zlippard:
Sounds good to me, will change the minor version to `0.8.0` and require the new map implementation of devices 👍

zlippard:
@runningcode Done deal!  I figured bumping to `0.8.0` made sense, if not, I can revert that back to `0.7.2`.  This PR is rebased on top of the other recent PR merged in.

buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
```diff
@@ -106,15 +106,19 @@ internal class YamlWriter {
...
 
   @VisibleForTesting
-  internal fun createDeviceString(devices: List<Device>): String = buildString {
+  internal fun createDeviceString(devices: List<Map<String, String?>>): String = buildString {
     appendln("  device:")
     for (device in devices) {
-      appendln("  - model: ${device.model}")
-      appendln("    version: ${device.version}")
-      device.orientation?.let {
+      val model = device["model"]
```
runningcode:
```suggestion
      val model = device["model"]!!
```
I believe `model` and `version` are required fields. Rather than printing `- model: null`, let's fail fast here.
Maybe we can write a test for this case too.

zlippard: Good call, will fix this 👍

zlippard: Alright, I added a `RequiredDeviceKeyMissingException` and throw that when "model" or "version" are `null`.

buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
```diff
@@ -106,15 +106,21 @@ internal class YamlWriter {
   }
 
   @VisibleForTesting
-  internal fun createDeviceString(devices: List<Device>): String = buildString {
+  internal fun createDeviceString(devices: List<Map<String, String?>>): String = buildString {
```
runningcode: About to merge, but I need to talk this out. :) Is it possible for the value to be `null` here? For example if we have `orientation` in the map, the value for `orientation` must be non-null. Same goes for `model`?

zlippard: You know what, that's a good point haha.  I'll remove the `?` that's unnecessary

zlippard: Fixed :) 8b6369454c00fadc2f9160668f089d8c7888f393

runningcode:
Thanks for helping out and fixing all those nitpicks. I've realized a lot of people are using this library so we want to make this API as nice as possible.

zlippard:
Happy to help!  It's a great plugin to use alongside Flank :)

PR #77, opened 2019-11-04
Support flank version 8.1.0

tahirhajizada:
Add new feature from flank version 8.1.0.
`keep-file-path`
Keeps the full path of downloaded files from a Google Cloud Storage bucket. Required when file names are not unique. Disabled by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,7 @@ fladle {
         }
     }
     resultsBucket("my-results-bucket-name")
+    keepFilePath = true
 }
 ```
 
@@ -172,6 +173,9 @@ Monitor and record performance metrics: CPU, memory, network usage, and FPS (gam
 
 ### resultsBucket
 The name of a Google Cloud Storage bucket where raw test results will be stored.
+
+### keepFilePath
+Keeps the full path of downloaded files from a Google Cloud Storage bucket. Required when file names are not unique. Disabled by default.
 
 ---
 ### Error APK file not found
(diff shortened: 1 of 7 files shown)


runningcode:
Hey, thanks for the contribution. Can we add some documentation to the readme around this flag and what it does?

tahirhajizada:
@runningcode Yes sure. I was going to add it.

tahirhajizada:
@runningcode I see there is already one open PR with version `0.7.2-SNAPSHOT`, in this case which version should I put?

README.md
```diff
@@ -173,6 +174,9 @@ Monitor and record performance metrics: CPU, memory, network usage, and FPS (gam
 ### resultsBucket
 The name of a Google Cloud Storage bucket where raw test results will be stored.
 
+### keepFilePath
+Keeps the full path of downloaded files from a Google Cloud Storage bucket. Required when file names are not unique. Disabled by default.
```
runningcode: what file names would not be unique? how does one know that one should enable this?

tahirhajizada: As an example. When the toggle is disabled downloaded files are stored at the root of device.
`results/2019-09-20-13-57-45.1454_nGys/shard_0/NexusLowRes-28-en-portrait/file.png`

But when it is enabled, then the full path of the file kept.
`results/2019-09-20-13-57-45.1454_nGys/shard_0/NexusLowRes-28-en-portrait/{package name}/file.png`

PR #83, opened 2019-11-12
Add support `additional-app-test-apks`

japplin:
Let me know if this dsl is acceptable if not, let me know how to adjust it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,7 @@ fladle {
     flankVersion("4.4.0")
     debugApk("$buildDir/outputs/apk/debug/sample-debug.apk")
     instrumentationApk("$buildDir/outputs/apk/androidTest/debug/sample-debug-androidTest.apk"
+    additionalTestApks = ["$buildDir/outputs/apk/debug/sample-debug.apk": ["$buildDir/outputs/apk/androidTest/debug/sample2-debug-androidTest.apk"]]
     autoGoogleLogin = true
     testShards = 5
     smartFlankGcsPath = "gs://tmp_flank/flank/test_app_android.xml"
@@ -120,6 +121,9 @@ This is the path to the app's debug apk.
 
 ### instrumentationApk
 This is the path to the app's instrumentation apk.
+
+### additionalTestApks
+Paths to additional test configurations.
 
 ### autoGoogleLogin
 Whether or not to automatically log in using a preconfigured google account. [More Info](https://cloud.google.com/sdk/gcloud/reference/firebase/test/android/run#--auto-google-login)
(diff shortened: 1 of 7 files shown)


runningcode:
Hey, thanks for the contribution! 
I think we learned that it's best not to use custom classes in the API of fladle. We removed `Device` from the API recently.
https://github.com/runningcode/fladle/pull/79

I'd like to not use a custom class here. I like where this is going though. Perhaps if this were a Map<String, String> I think it would work.

japplin:
Hey thanks for the update, what do you think of using Map<String, List<String>>? It would help avoid redeclaring the same debugApk.

runningcode:
Sounds good!

japplin:
Updated

runningcode:
Hey, sorry, I must have missed this in the first review. I was looking at this from mobile. I would like to be able to keep the current api with `debugApk` and `instrumentationApk`. Would it be possible to use this map only for the `additional-test-apks`?

japplin:
I've updated again with restored debugApk and instrumentationApk fields

buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
```diff
@@ -56,6 +56,17 @@ internal class YamlWriter {
         appendln("  - $file")
       }
     }
+    val testApks = config.additionalTestApks.flatMap { (debugApk, instrumentationApks) ->
```
runningcode: This code can be somewhat simplified a bit since the instrumentation apk will take whatever value is above is as the debug apk. See the readme. https://github.com/TestArmada/flank/blob/master/README.md
```
if (config.additionalTestApks.isNotEmpty()) {
  appendln("  additional-app-test-apks:")
  config.additionalTestApks.entries.forEach { debugApk, instrumentationApks ->
    appendln("    - app: ${debugApk}")
    instrumentationApks.forEach {
        appendln("    test: ${it}")
    }
  }
```
On the other hand, I do like the more explicit output format that you created here. Let me think this one over overnight. :)

buildSrc/src/main/java/com/osacky/flank/gradle/FlankGradlePlugin.kt
```diff
@@ -104,6 +104,9 @@ class FlankGradlePlugin : Plugin<Project> {
     }
     checkNotNull(base.debugApk!!) { "debugApk file cannot be null ${base.debugApk}" }
     checkNotNull(base.instrumentationApk!!) { "instrumentationApk file cannot be null ${base.instrumentationApk}" }
+    base.additionalTestApks.forEach {
```
runningcode: 👍

README.md
```diff
@@ -121,6 +122,9 @@ This is the path to the app's debug apk.
 ### instrumentationApk
 This is the path to the app's instrumentation apk.
 
+### additionalTestApks
```
runningcode: can you add a short sample? and that it is a map of debug apks to a list of instrumentation apks? thanks :)

buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
japplin: Should I change this?

PR #116, opened 2020-05-14
Use ListProvider for additionalTestApks

### LINKED ISSUE #114: Change API for additionalTestApks
The current API for additional test apks uses a map
```
additionalTestApks  = [ 
          // null in order to use existing app apk. key order matters
          null : [ "foo/bar/room-test.apk" ],
         ("${rootProject.build}/foo/bar/sample-app.apk".toString()) : [ "foo/bar/sample-app-test.apk", "foo/bar/room-test-3.apk" ]
         ("${rootProject.build}/module/foo/app.apk".toString()) : ["orange/blue/test-apk.apk"],
         ("${findApkInApp(buildType)}".toString()) : [ "red/green/test2-apk"]
]
```
The issue with the map is the weird groovy syntax since gstrings can't be keys.
"${rootProject.build}/module/foo/app.apk" or "${findApkInApp(buildType)}" would normally only be evaluated when you call toString . But since this is a map, you need to resolve the string in order to create the map's key.
That means that the `findApkInApp` function needs to be evaluated at config time in order to create the map's key.

I propose changing the syntax to the following:

List option:
```
additionalTestApks  = [ 
          "test: foo/bar/room-test.apk",
          "app: ${rootProject.build}/foo/bar/sample-app.apk",
          "test: foo/bar/sample-app-test.apk",
          "test: foo/bar/room-test-3.apk",
          "app: ${rootProject.build}/module/foo/app.apk",
          "test: orange/blue/test-apk",
          "app: ${findApkInApp(buildType)}",
          "test: red/green/test2-apk"
]
```

Both examples should result in the same tests being run in the same order, but the second list option seems to have much cleaner syntax.

runningcode:
Fixes #114

### DIFF
--- a/buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
+++ b/buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
@@ -61,15 +61,12 @@ internal class YamlWriter {
         appendln("  - $file")
       }
     }
-    val testApks = config.additionalTestApks.flatMap { (debugApk, instrumentationApks) ->
-      instrumentationApks.map { debugApk to it }
-    }
+    val additionalTestApks = config.additionalTestApks.getOrElse(emptyList())
 
-    if (testApks.isNotEmpty()) {
+    if (additionalTestApks.isNotEmpty()) {
       appendln("  additional-app-test-apks:")
-      testApks.forEach {
-        appendln("    - app: ${it.first}")
-        appendln("      test: ${it.second}")
+      additionalTestApks.forEach {
+        appendln("    - $it")
       }
     }
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -117,7 +117,31 @@ This is the path to the app's instrumentation apk. Supports wildcard characters.
 
 ### additionalTestApks
 Paths to additional test configurations.
+Order matters. A test apk is run with the nearest previous listed app apk.
+It is not required to list an app apk here. If there is no app apk listed in additionalTestApks, the test apks are run against the [#debugApk].
 
+=== "Groovy"
+    ```groovy
+    additionalTestApks.value(project.provider { [
+    "app: ../main/app/build/output/apk/debug/app.apk",
+    "test: ../main/app/build/output/apk/androidTest/debug/app-test.apk",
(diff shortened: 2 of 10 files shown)


buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
```diff
@@ -61,15 +61,12 @@ internal class YamlWriter {
...
+    val additionalTestApks = config.additionalTestApks.getOrElse(emptyList())
 
-    if (testApks.isNotEmpty()) {
+    if (additionalTestApks.isNotEmpty()) {
       appendln("  additional-app-test-apks:")
-      testApks.forEach {
-        appendln("    - app: ${it.first}")
-        appendln("      test: ${it.second}")
+      additionalTestApks.forEach {
+        appendln("    - $it")
```
rciovati, line 69: Reading flank's doc it seems that the `test` apk is not always expected to have a dash at the beginning. I'm not sure why, but maybe it's something that could be problematic here?

```
  ## Include additional app/test apk pairs in the run. Apks are unique by just filename and not by path!
  ## If app is omitted, then the top level app is used for that pair.
  # additional-app-test-apks:
  #  - app: ../test_app/apks/app-debug.apk
  #    test: ../test_app/apks/app1-debug-androidTest.apk
  #  - test: ../test_app/apks/app2-debug-androidTest.apk
```

runningcode: We discussed this in the firebase slack. It seems there is no difference: https://firebase-community.slack.com/archives/C72V6UW8M/p1589445650211000

rciovati: Cool, good to know! 👍

docs/configuration.md
```diff
@@ -118,6 +118,28 @@ This is the path to the app's instrumentation apk. Supports wildcard characters.
 ### additionalTestApks
 Paths to additional test configurations.
```
julioz, line 119: maybe it's a good chance to expand a bit on the relationship between `app` and `test` entries, and most importantly mention the "ordering aspect" of this list.

PR #126, opened 2020-05-20
Lazy configuration

pawelpasterz:
Implement lazy configuration for all properties


buildSrc/src/main/java/com/osacky/flank/gradle/FlankGradleExtension.kt
```diff
@@ -6,33 +6,35 @@ import org.gradle.api.NamedDomainObjectContainer
...
   // Project id is automatically discovered by default. Use this to override the project id.
-  override var projectId: String? = null
+  override var projectId: Property<String> = objects.property()
   override val serviceAccountCredentials: RegularFileProperty = objects.fileProperty()
-  override var useOrchestrator: Boolean = false
-  override var autoGoogleLogin: Boolean = false
-  override var devices: List<Map<String, String>> = listOf(mapOf("model" to "NexusLowRes", "version" to "28"))
+  override var useOrchestrator: Property<Boolean> = objects.property()
+  override var autoGoogleLogin: Property<Boolean> = objects.property()
+  override var devices: ListProperty<Map<String, String>> = objects.listProperty()
```
runningcode, line 23: can we specify the default here using convention? this ensure that no matter where it is accessed, the same default value is used.
```suggestion
  override var devices: ListProperty<Map<String, String>> = objects.listProperty().convention(listOf(mapOf("model" to "NexusLowRes", "version" to "28"))
```

buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
```diff
@@ -24,18 +24,18 @@ internal class YamlWriter {
   }
 
   internal fun writeFlankProperties(config: FladleConfig): String = buildString {
-    val testShards = config.testShards
-    val shardTime = config.shardTime
-    val repeatTests = config.repeatTests
-    val smartFlankGcsPath = config.smartFlankGcsPath
-    val filesToDownload = config.filesToDownload
-    val projectId = config.projectId
+    val testShards = config.testShards.orNull
```
runningcode, line 27: i think it might be more idiomatic to check `testShards.isPresent` on line 42.
i have a preference against using `null` to represent a state or value. :)

buildSrc/src/test/java/com/osacky/flank/gradle/YamlWriterTest.kt
```diff
@@ -355,64 +352,64 @@ class YamlWriterTest {
...
   fun writeTestTargetsAndResultsHistoryName() {
     val extension = emptyExtension {
-      resultsHistoryName = "androidtest"
-      testTargets = listOf("class com.example.Foo")
+      resultsHistoryName.set("androidtest")
+      testTargets.set(listOf("class com.example.Foo"))
     }
 
-    assertEquals("  use-orchestrator: false\n" +
+    assertEquals("  results-history-name: androidtest\n" +
```
runningcode, line 407: can we keep the ordering the same here?

pawelpasterz: Ok, I can rearrange that. This change is result of my refactoring in `YamlWriter` class where my intention was to keep related code in one place, example -- all logics related to default values is grouped together like:
```
// without default values
    val testShards = config.testShards
    val shardTime = config.shardTime
    val repeatTests = config.repeatTests
    val smartFlankGcsPath = config.smartFlankGcsPath
    val projectId = config.projectId
    val runTimeout = config.runTimeout
    val localResultsDir = config.localResultsDir

    if (testShards.isPresent) appendln("  max-test-shards: ${testShards.get()}")
    if (shardTime.isPresent) appendln("  shard-time: ${shardTime.get()}")
    if (repeatTests.isPresent) appendln(repeatTestsLine(repeatTests.get()))
    if (smartFlankGcsPath.isPresent) appendln("  smart-flank-gcs-path: ${smartFlankGcsPath.get()}")
    if (projectId.isPresent) appendln("  project: ${projectId.get()}")
    if (runTimeout.isPresent) appendln("  run-timeout: ${runTimeout.get()}")
    if (localResultsDir.isPresent) appendln("  local-result-dir: ${localResultsDir.get()}")
```

I think the best would be to refactor tests to check for presents for specific lines instead of whole, hardcoded string. What do you think?

bootstraponline:
hey, what's the status of this pull request?

pawelpasterz:
@bootstraponline 
There is a problem with additional configs implemented with this approach, since config parameters are `Property` objects (not primitive values) each config (additional and 'main') ends up with reference to the same object, which leads to overwritten configuration. That needs to be solved in order to move forward

cc
@runningcode

buildSrc/src/main/java/com/osacky/flank/gradle/FladleConfigImpl.kt
```diff
@@ -2,43 +2,44 @@ package com.osacky.flank.gradle
 
 import org.gradle.api.file.RegularFileProperty
 import org.gradle.api.provider.ListProperty
+import org.gradle.api.provider.MapProperty
 import org.gradle.api.provider.Property
 
 data class FladleConfigImpl(
   internal val name: String,
-  override var projectId: String?,
+  override var projectId: Property<String>,
```
jan-gogo, line 10: I am not very familiar with gradle api but if properties are mutable under the  hood you should consider to change `var` -> `val`.

runningcode:
Sorry this took me forever to look at. You can use this pattern to prevent the values from being overridden in other configurations: https://github.com/runningcode/fladle/pull/169/files#diff-670e340c5b646cd261359b8466643022R110

pawelpasterz:
@runningcode I am going to close this PR. It's already far behind the master branch an it's extremely difficult to resolve conflict and ensure it's working correctly at the same time. I'll implement this in separate PR

runningcode:
Thanks for the update!

Issue #136, opened 2020-06-03
timeoutMin should be timeout with configuration of units


digitalbuddha:
Howdy! I have a request. I'd like to run robo-tests for 30s. Currently fladle only allows timeoutMin while flank allows timeout where I can set something like 30s.  Could we add that api to fladle as well. Thank you

runningcode:
I honestly have no idea what the difference is between 
https://runningcode.github.io/fladle/configuration/#runtimeout and https://runningcode.github.io/fladle/configuration/#timeoutmin

@pawelpasterz can you shed some light here? What is the difference between those two values and what @digitalbuddha is asking for?

Also @digitalbuddha can you link the flank docs about this feature just so we're clear which one we're talking about?

pawelpasterz:
[1] `runTimeout` -- time after flank cancels run. Before exiting process flank cancels all running matrices if there are any. It's flank's feature.

[2] `timeoutMin` --  time after which FTL cancels running tests/matrices. It's FTL's feature

[2] is send as parameter to FTL during matrix construction while [1] is flank's internal.

Let me know if my explanation is clear :)

runningcode:
Yes, thank you very much! This would be a good opportunity for documentation improvements. :)

digitalbuddha:
Hi friends, refer to this thread https://firebase-community.slack.com/archives/C1MTSQ5QT/p1591219224240200?thread_ts=1591218666.237500&cid=C1MTSQ5QT my understanding is that I should be able to set timeout as seconds and not just minutes.

pawelpasterz:
Correct, flank supports different units. I think in fladle minutes are hardcoded
https://github.com/runningcode/fladle/blob/5ba419507ac4f2e694dde60431b9f6acd9260d32/buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt#L103

runningcode:
oh nice find. yes please do!

runningcode:
Was released in 0.10.2

PR #137, opened 2020-06-05
Add time units support for timeout

pawelpasterz:
Feature request #136

### DIFF
--- a/buildSrc/src/main/java/com/osacky/flank/gradle/FladleConfig.kt
+++ b/buildSrc/src/main/java/com/osacky/flank/gradle/FladleConfig.kt
@@ -160,5 +160,5 @@ interface FladleConfig {
    * * 200s -> 200 seconds
    * * 100  -> 100 seconds
    */
-  var timeout: String
+  var testTimeout: String
 }
--- a/buildSrc/src/main/java/com/osacky/flank/gradle/FladleConfig.kt
+++ b/buildSrc/src/main/java/com/osacky/flank/gradle/FladleConfig.kt
@@ -31,8 +31,6 @@ interface FladleConfig {
 
   var resultsHistoryName: String?
 
-  var timeoutMin: Int
-
   var directoriesToPull: List<String>
 
   var filesToDownload: List<String>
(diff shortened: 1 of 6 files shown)


pawelpasterz:
Hey @runningcode , please tak a look, thanks!

buildSrc/src/main/java/com/osacky/flank/gradle/FladleConfig.kt
```diff
@@ -149,4 +147,18 @@ interface FladleConfig {
...
+   * The maximum possible testing time is 45m on physical devices and 60m on virtual devices.
+   * The TIMEOUT units can be h, m, or s. If no unit is given, seconds are assumed.
+   *
+   * Examples:
+   * * 1h   -> 1 hour
+   * * 5m   -> 5 minutes
+   * * 200s -> 200 seconds
+   * * 100  -> 100 seconds
+   */
+  var timeout: String
```
runningcode, line 163: can we call this `testTimeout` in order to distinguish it from the `runTimeout`?
also if we're going to break the API, can we break it all the way and make this a property while we're at it? thanks!

pawelpasterz: No problem, I'll rename it, thanks!
Regarding property, I think it will break additional config feature in fladle. It's how they are created, and with `Property`, its reference will be assigned, not value and that  may lead to overwriting configs.

That's why I wanted to keep this change separated from that one.

bootstraponline:
hey, what's the status of this pull request?

pawelpasterz:
@bootstraponline 
Since `Property` problem for additional configs is not solved yet (https://github.com/runningcode/fladle/pull/126#issuecomment-645754679) I am not sure if I should attach this to PR scope. See my comment above. But it's up to @runningcode :)

runningcode:
Sorry for taking so long, yeah let's merge this for now and figure out how to fix the additional configs problem later.

runningcode:
Thank you very much for taking care of this @pawelpasterz

PR #142, opened 2020-07-08
Support output-style property

### LINKED ISSUE #140: Duplicate status messages
**Describe the bug**
Running flank using fladle gives duplicated messages on output. 

**Details**
Flank provides 3 different types of output style [single, multi, verbose]. `--output-style=multi` is default when flank is running more than one execution. It is using ANSI codes to overwrite multiple lines in output. Some consoles may not supports ANSI codes so it can results with duplicated messages. Also running flank through gradle may give same results.

**To reproduce**
Run fladle with `max-test-shards`  > 1. It's important to have more than one shard. The expected result was pasted in issue https://github.com/Flank/flank/issues/841.

**Additional context**
* `--output-style=multi` displays separated status for each shard execution in separated line, lines are updated over time. Similar to gradle output when running multiple tasks in parallel. Requires ANSI codes support.
* `--output-style=single` displays shortened status of all executions in single line. Similar to gcloud output when running with sharding. Should work on any console.

**Possible solutions**
* Add option `output-style` to fladle so users can specify which one they want.
* Add `output-style=single` to executions under the hood so all executions on fladle will display single output.

francescocervone:
Closes #140

### DIFF
--- a/buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
+++ b/buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
@@ -44,7 +44,6 @@ internal class YamlWriter {
     val smartFlankDisableUpload = config.smartFlankDisableUpload
     val localResultsDir = config.localResultsDir.orNull
     val testTargetsAlwaysRun = config.testTargetsAlwaysRun
-    val outputStyle = config.outputStyle.orNull
 
     appendln("flank:")
 
@@ -95,9 +94,7 @@ internal class YamlWriter {
       }
     }
 
-    outputStyle?.let {
-      appendln("  output-style: $it")
-    }
+    appendln("  output-style: ${config.outputStyle.get()}")
   }
 
   internal fun writeAdditionalProperties(config: FladleConfig): String = buildString {
--- a/buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
+++ b/buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
@@ -44,7 +44,7 @@ internal class YamlWriter {
     val smartFlankDisableUpload = config.smartFlankDisableUpload
     val localResultsDir = config.localResultsDir.orNull
     val testTargetsAlwaysRun = config.testTargetsAlwaysRun
-    val outputStyle = config.outputStyle
+    val outputStyle = config.outputStyle.orNull
 
     appendln("flank:")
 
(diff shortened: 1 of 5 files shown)


runningcode:
Thank you for the contribution! Would you be able to change the string to use [Lazy Properties](https://docs.gradle.org/current/userguide/lazy_configuration.html#creating_property_provider)

Should we be defaulting to using single as the default? I would assume that most flank use cases run more than one shard. Is this broken for all cases using more than one shard?

francescocervone:
@runningcode you are welcome!
>Would you be able to change the string to use Lazy Properties?

Of course.

>Should we be defaulting to using single as the default?

I didn't because it looked weird to me that Fladle overrides some Flank's property's default. As a Fladle user, I'd prefer to default output style to `single`. It's your call.

>Is this broken for all cases using more than one shard?

What do you mean when you say "for all cases"? I tried with just 2 shards and with 50 shards. The issue occurs in both cases.

runningcode:
Thanks for using the lazy property!

> What do you mean when you say "for all cases"? I tried with just 2 shards and with 50 shards. The issue occurs in both cases.

I mean if it improves the experience for 2 shards and 50 shards, I think we should make it the default? The only case where it degrades the experience is for 1 shard? Is that true?

francescocervone:
Yes, as far as I understand.

buildSrc/src/test/java/com/osacky/flank/gradle/YamlWriterTest.kt
```diff
@@ -790,6 +807,22 @@ class YamlWriterTest {
     assertTrue(properties.contains("  num-uniform-shards: 20"))
   }
 
+  @Test
+  fun writeOutputStyle() {
+    val properties = emptyExtension {
+      outputStyle.set("anyString")
```
runningcode, line 813: i guess this will fail when the flank command is executed?
i'm fine with that, just wanted to point it out. i don't think validating this option on the fladle side makes sense in case more options are introduced by flank down the line.

buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
```diff
@@ -44,6 +44,7 @@ internal class YamlWriter {
     val smartFlankDisableUpload = config.smartFlankDisableUpload
     val localResultsDir = config.localResultsDir.orNull
     val testTargetsAlwaysRun = config.testTargetsAlwaysRun
+    val outputStyle = config.outputStyle.orNull
```
runningcode, line 47: i think `orNull` isn't necessary or has no effect here since the convention is set.

buildSrc/src/test/java/com/osacky/flank/gradle/YamlWriterTest.kt
francescocervone: Yes, I did that on purpose for the same reason you pointed out.

Issue #147, opened 2020-07-11
Write Yaml file to task specific output directory


runningcode:
In order to support running multiple simultaneous flank configs, the first step is to write the yaml output file to a task specific directory.
https://github.com/runningcode/fladle/blob/master/buildSrc/src/main/java/com/osacky/flank/gradle/YamlConfigWriterTask.kt#L28

Running the flank command will also need to reference this specific yaml file: https://github.com/runningcode/fladle/blob/master/buildSrc/src/main/java/com/osacky/flank/gradle/FladlePluginDelegate.kt#L90

See: https://github.com/runningcode/fladle/issues/141

PR #159, opened 2020-08-14
Declare YamlConfigWriterTask output so it can be used from other tasks.

CristianGM:
I wanted to declare the inputs to (so it would be able to check if it's up-to-date) but:
`Caused by: org.gradle.internal.snapshot.ValueSnapshottingException: Could not serialize value of type FlankGradleExtension`

In any case this is useful so the yaml can be used as an input for FlankExecutionTask (one of many)

### DIFF
--- a/buildSrc/src/main/java/com/osacky/flank/gradle/YamlConfigWriterTask.kt
+++ b/buildSrc/src/main/java/com/osacky/flank/gradle/YamlConfigWriterTask.kt
@@ -3,12 +3,19 @@ package com.osacky.flank.gradle
 import org.gradle.api.DefaultTask
 import org.gradle.api.file.ProjectLayout
 import org.gradle.api.tasks.Internal
+import org.gradle.api.tasks.OutputFile
 import org.gradle.api.tasks.TaskAction
+import java.io.File
 import javax.inject.Inject
 
-open class YamlConfigWriterTask @Inject constructor(private val config: FladleConfig, private val extension: FlankGradleExtension, private val projectLayout: ProjectLayout) : DefaultTask() {
+open class YamlConfigWriterTask @Inject constructor(private val config: FladleConfig, private val extension: FlankGradleExtension, projectLayout: ProjectLayout) : DefaultTask() {
 
   private val yamlWriter = YamlWriter()
+
+  private val fladleDir = projectLayout.fladleDir.get().asFile
+
+  @OutputFile
+  val fladleConfigFile: File = fladleDir.resolve("flank.yml")
 
   @Internal
   override fun getDescription(): String {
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/java/com/osacky/flank/gradle/YamlConfigWriterTask.kt
```diff
@@ -3,13 +3,20 @@ package com.osacky.flank.gradle
...
 
-open class YamlConfigWriterTask @Inject constructor(private val config: FladleConfig, private val extension: FlankGradleExtension, private val projectLayout: ProjectLayout) : DefaultTask() {
+open class YamlConfigWriterTask @Inject constructor(private val config: FladleConfig, private val extension: FlankGradleExtension, projectLayout: ProjectLayout) : DefaultTask() {
 
   private val yamlWriter = YamlWriter()
 
+  private val fladleDir = projectLayout.fladleDir.get().asFile
+
+  @OutputFile
+  val fladleConfigFile: File = fladleDir.resolve("flank.yml")
```
runningcode, line 18: is it possible to make this simply a property so that it can be lazily evaluated?

buildSrc/src/main/java/com/osacky/flank/gradle/YamlConfigWriterTask.kt
```diff
@@ -3,13 +3,20 @@ package com.osacky.flank.gradle
...
 import org.gradle.api.tasks.TaskAction
+import java.io.File
 import javax.inject.Inject
 
-open class YamlConfigWriterTask @Inject constructor(private val config: FladleConfig, private val extension: FlankGradleExtension, private val projectLayout: ProjectLayout) : DefaultTask() {
+open class YamlConfigWriterTask @Inject constructor(private val config: FladleConfig, private val extension: FlankGradleExtension, projectLayout: ProjectLayout) : DefaultTask() {
 
   private val yamlWriter = YamlWriter()
 
+  private val fladleDir = projectLayout.fladleDir.get().asFile
```
runningcode, line 15: Not necessary for this PR, just something to be aware of this will cause overlapping outputs when using multiple configs. I wanted to fix this with: https://github.com/runningcode/fladle/issues/147
We could create a directory based on the config name to solve this.

runningcode: does this need to be annotated with `@Internal`? (check if there are build warnings around this)

CristianGM: `@Internal` is not needed because private fields don't generate getters.

I usually make the output configurable, but I'm trying to do the minimum changes to the current behavior or I would make a very big PR.

I like to do something like:
```
@get:OutputDirectory
val outputDir: DirectoryProperty = objectFactory.directoryProperty().convention(layout.buildDirectory.dir("fladle"))
```

buildSrc/src/main/java/com/osacky/flank/gradle/YamlConfigWriterTask.kt
CristianGM: Isn't creating a File almost free? It's only when you try to access it that it involves the IO.
I can make it lazy but I don't see the point (maybe I'm missing something)

runningcode: "almost" :D

buildSrc/src/main/java/com/osacky/flank/gradle/YamlConfigWriterTask.kt
runningcode: yeah, i like your approach of making this minimal, we can always fix the other stuff later. thanks!

PR #160, opened 2020-08-14
Increase Gradle min version to 6.0

CristianGM:
Thats required to make FlankExecutionTask a custom task (not extending JavaExec) so we can declare inputs and don't need to set all the properties in the init block.

### DIFF
--- a/fladle-plugin/src/main/java/com/osacky/flank/gradle/FladlePluginDelegate.kt
+++ b/fladle-plugin/src/main/java/com/osacky/flank/gradle/FladlePluginDelegate.kt
@@ -35,7 +35,7 @@ class FladlePluginDelegate {
   }
 
   private fun checkMinimumGradleVersion() {
-    // Gradle 4.9 is required because we use the lazy task configuration API.
+    // Gradle 6.0 is required because we use ExecOperations.
     if (GRADLE_MIN_VERSION > GradleVersion.current()) {
       throw GradleException("Fladle requires at minimum version $GRADLE_MIN_VERSION. Detected version ${GradleVersion.current()}.")
     }
@@ -195,7 +195,7 @@ class FladlePluginDelegate {
     get() = configurations.getByName(FLADLE_CONFIG)
 
   companion object {
-    val GRADLE_MIN_VERSION: GradleVersion = GradleVersion.version("5.5")
+    val GRADLE_MIN_VERSION: GradleVersion = GradleVersion.version("6.0")
     const val TASK_GROUP = "fladle"
     const val FLADLE_CONFIG = "fladle"
     fun Project.log(message: String) {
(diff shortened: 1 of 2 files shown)


runningcode:
Hey, thanks for the idea. It is possible to declare inputs on `FlankExecutionTask`. It is also possible to not set all the properties in the init block.
Can you describe what limitations you faced with these two things?

CristianGM:
How can you set the arguments after the init when using JavaExec?
- overriding the doFirst? It may be unintuitive for consumers of the plugin in case they want to add a doFirst themselves.
- adding a doFirst inside the init would be the only way.

Anyway the reason is to have a task that you can control, my next PR will have something like:
```
  @TaskAction
  fun run() {
    if (serviceAccountCredentials.isPresent) {
      check(serviceAccountCredentials.get().asFile.exists()) { "serviceAccountCredential file doesn't exist ${serviceAccountCredentials.get()}" }
    }

    execOperations.javaexec {
      classpath = fladleConfig
      args = if (dumpShards.get()) {
        listOf("firebase", "test", "android", "run", "--dump-shards")
      } else {
        listOf("firebase", "test", "android", "run")
      }

      if (serviceAccountCredentials.isPresent) {
        environment(mapOf("GOOGLE_APPLICATION_CREDENTIALS" to serviceAccountCredentials.get()))
      }

      main = "ftl.Main"
      workingDir(projectLayout.fladleDir)
    }.assertNormalExitValue()
  }
```

No `config` as constructor parameter, just declared inputs
Not everything set from the plugin (delegate), easier to reason about because the task has a real behavior, right now you don't need to have a custom Task type, you could directly use JavaExec from the plugin.

..and if we decide at some point to use flank as a library instead of as an executable someday it will be times simpler, but that's just an imaginary use case for now.

I wouldn't like to write all of this inside the init -> doFirst

runningcode:
Ok, i think im convinced of doing the 6.0 update. But let me cut a release before merging this.
If anyone else needs fladle compatibility with 5.5, now's your chance to comment!

bootstraponline:
@runningcode are we good to release and merge?

bootstraponline:
@runningcode 🙂

runningcode:
Hi @bootstraponline I appreciate your reminder here. I wrote above that I was convinced of a 6.0 update. I will do it when I see the PR that requires it _and_ there is *no* possibility to maintain backward compatibility. Please feel free to submit PRs that use features requiring Gradle 6.0.

It seems that the main argument for raising the minimum Gradle version is that it is the only way to perform the following:

> How can you set the arguments after the init when using JavaExec?

I am not sure I understand why we would want to do this.
 
>Not everything set from the plugin (delegate), easier to reason about because the task has a real behavior, right now you don't need to have a custom Task type, you could directly use JavaExec from the plugin.

There is definitely a benefit to defining a custom task and passing the parameters to the custom task. It helps keep the logic separate from the plugin class.

PR #161, opened 2020-08-14
Write Yaml file to task specific output directory

### LINKED ISSUE #147: Write Yaml file to task specific output directory

CristianGM:
Fixes #147

I know there are some duplicates, but even worst I'm doing a lot in configuration phase, with the change to Gradle 6.0 and using ExecOperations I had that done in a better way, only reading values at execution phase.


buildSrc/src/main/java/com/osacky/flank/gradle/YamlConfigWriterTask.kt
```diff
@@ -2,20 +2,21 @@ package com.osacky.flank.gradle
...
+open class YamlConfigWriterTask @Inject constructor(private val config: FladleConfig, private val extension: FlankGradleExtension, projectLayout: ProjectLayout, objectFactory: ObjectFactory) : DefaultTask() {
 
   private val yamlWriter = YamlWriter()
 
-  private val fladleDir = projectLayout.fladleDir.get().asFile
-
   @OutputFile
-  val fladleConfigFile: File = fladleDir.resolve("flank.yml")
+  val flankConfig: Property<RegularFile> =
+    objectFactory.fileProperty().convention(projectLayout.fladleDir.map { it.file("flank.yml") })
```
runningcode, line 19: what does `projectLayout.fladleDir.map { it.file("flank.yml") }` return? can we just use that?

CristianGM: That would be a Provider, not a Property

runningcode: Ah, I just took a look at this and I think I understand it better now. Is it possible to only pass the string of the config name as an input to this task and then that can be used to compute the the output directory?
Otherwise we are only using that name as an output. That is, two tasks with the same inputs will have different output directories.

CristianGM: What's the issue with that?
What we need is to declare the inputs. Tight now it's not possible because the extension is not serializable but also, I don't think the extension is the right data to provide to the yaml writer.

Let me move config to an input, so at least we'll fix what you are asking for.

runningcode (approved):

runningcode:
Thanks so much for this contribution!

buildSrc/src/main/java/com/osacky/flank/gradle/FladlePluginDelegate.kt
```diff
@@ -57,20 +57,25 @@ class FladlePluginDelegate {
   }
 
   private fun TaskContainer.createTasksForConfig(base: FlankGradleExtension, config: FladleConfig, project: Project, name: String) {
+    val writeConfigProps = register("writeConfigProps$name", YamlConfigWriterTask::class.java, base)
```
runningcode, line 60: @CristianGM curious, why did you remove passing the `config` in the constructor here and instead set it during the configuration?

CristianGM: can I declare a Task input as a constructor parameter?

CristianGM: I guess I could just add a getter to declare the input.
But actually is that I never used constructor parameters for a task (unless for injected services) and the documentation doesn't mention them a lot.

runningcode: Yeah, like this: https://github.com/runningcode/fladle/pull/162/files#diff-0a17c2db8a8d3d5a093b29f3284b93b5R16

PR #162, opened 2020-08-18
Create file lazily.

runningcode:
This avoids creating the file object until as late as possible. This is needed in case the projectlayout changes in the configuration phase.

This also fixes build deprecations added in previous pull requests.


runningcode:
@CristianGM could you take a look here? I incorporated your changes with mine.

Issue #163, opened 2020-08-24
Use latest stable version of Flank by default


bootstraponline:
Flank `flankVersion = "20.07.0"` has some bugs related to outcome that have been resolved in `v20.08.3`

bootstraponline:
https://github.com/runningcode/fladle/commit/bc5e1f312a8de984c3805c04ac4d648695ab9ef5

PR #164, opened 2020-08-24
#163 Update Flank to 20.08.3

piotradamczyk5:
Update Flank to `v20.08.3`


runningcode (approved):

piotradamczyk5:
hey @runningcode CircleCi build fails seems to be an issue of CI itself, do you know any workaround for it?

CristianGM:
@piotradamczyk5 
> exit value 137
it means we are consuming more memory than what is available, maybe it's because the usual error (no idea how CircleCI works, but I guess it's similar to Travis) where the JVM tells gradle we have more memory (all the memory the real machine has) but the Docker is only allowed to consume a limited amount, in this case 4GB and 2 CPUs)

Maybe something as simple as a restart helps, but it will be hiding the real issue.

I don't know if that happened before with Fladle but if it didn't... does Flank consume more memory now than the previous version?

piotradamczyk5:
> @piotradamczyk5
> 
> > exit value 137
> 
> it means we are consuming more memory than what is available, maybe it's because the usual error (no idea how CircleCI works, but I guess it's similar to Travis) where the JVM tells gradle we have more memory (all the memory the real machine has) but the Docker is only allowed to consume a limited amount, in this case 4GB and 2 CPUs)
> 
> Maybe something as simple as a restart helps, but it will be hiding the real issue.
> 
> I don't know if that happened before with Fladle but if it didn't... does Flank consume more memory now than the previous version?

I don't think that memory consumption increased significantly. I do not have any permission even to re-run the job, so I do not have any idea how to handle this issue. Maybe I will try creating another PR

piotradamczyk5:
> > @piotradamczyk5
> > > exit value 137
> > 
> > 
> > it means we are consuming more memory than what is available, maybe it's because the usual error (no idea how CircleCI works, but I guess it's similar to Travis) where the JVM tells gradle we have more memory (all the memory the real machine has) but the Docker is only allowed to consume a limited amount, in this case 4GB and 2 CPUs)
> > Maybe something as simple as a restart helps, but it will be hiding the real issue.
> > I don't know if that happened before with Fladle but if it didn't... does Flank consume more memory now than the previous version?
> 
> I don't think that memory consumption increased significantly.
> I am sure that Flank does not consume 4GB.
> I do not have any permission even to re-run the job, so I do not have any idea how to handle this issue.

Based on @pawelpasterz research Flank consumes around 0,64GB

runningcode:
I switched the repo to github actions to fix the memory issues. Can you rebase from master to see if this fixes the problems?

piotradamczyk5:
> I switched the repo to github actions to fix the memory issues. Can you rebase from master to see if this fixes the problems?

It solves the issue. Thank you!

Issue #92, opened 2020-04-23
Use lazy configuration and lazy properties for configuring Fladle


runningcode:
This will ensure that we don't resolve files at configuration time.

https://docs.gradle.org/current/userguide/lazy_configuration.html

runningcode:
example here: https://github.com/runningcode/fladle/commit/b47fbc159ea8df380eb131bb0e85a510461913e5

Issue #165, opened 2020-08-24
Support for Robo Tests without Robo script


ccjernigan:
This is an enhancement request to support robo tests without a robo script.  Firebase Test Lab supports this, although it appears fladle does not. In the fladle source, the decision to do a robo test is based on the roboscript file path being non-null.

Issues sometimes come up with R8, ProGuard, or DexGuard that make robo testing the my release APK helpful in catching issues that might occur during first launch.  However, DexGuard also obfuscates resource IDs which makes creating a roboscript for a release build much more difficult.

Being able to launch a robo test without a roboscript will enable my use case, but would also lower the barrier for others who want to run a robo test without recording a script.

runningcode:
Oh, I didn't realize this was a use case. Thanks for reporting. Do you know if flank itself supports this?

runningcode:
Yes, flank supports this but you must specify `robo-directives` is that your use case?

ccjernigan:
I was initially looking to effectively simulate prelaunch reports, without having to upload to Google Play.  I've tried running this directly with the gcloud command without roboscript or robo-directives, which works.

Let me check on what's involved with robo-directives, because maybe an empty set of robo-directives would work.

runningcode:
Is the fix in fladle to allow a roboDirective to be specified without specifying a roboScript or an instrumentation APK? I can do this relatively quickly if this is what is needed. Please confirm. Thanks!

ccjernigan:
I tried with flank, and it looks like it has a requirement that one of these be set:

```
java -jar flank.jar firebase test android run
version: v20.08.4
revision: 0b328fa775ebc69980b765b35a654089a6235d10

One of following options must be specified [test, robo-directives, robo-script].
```

I tried just setting fake robo-directives but it still would not run, so this looks like a limitation in flank currently.

runningcode:
Ah, in that case, can you open an issue in the flank repository? Link it from here. It's likely we'll both need to make changes.

ccjernigan:
https://github.com/Flank/flank/issues/1084

ccjernigan:
FYI the underlying Flank issue was just resolved 🎉

runningcode:
Please do! Thanks!

PR #176, opened 2020-09-23
Implement lazy properties

pawelpasterz:
This will fix https://github.com/runningcode/fladle/issues/92

What was done?
1) Implement all remaining properties to use lazy API.
2) small refactor

What was not done?
1) there is a test failing (locally) `ConfigurationCacheTest#runFlank` (it is also falling on the master branch)
2) update with latest flank's command
3) enable to run robo test without robo script
4) refactor of `YamlWriterTest`

I decided to left above as they are since the scope of this PR is rather big


pawelpasterz:
Hey, @runningcode please take a look, thanks!

build.gradle
```diff
@@ -16,6 +16,7 @@ plugins {
     id "com.github.ben-manes.versions" version "0.30.0"
     id "com.osacky.fulladle"
     id "org.jmailen.kotlinter" version "3.0.2"
+    id "org.jetbrains.kotlin.jvm" version "$kotlin_version"
```
runningcode, line 19: can you tell me more about this change? why do we need this?

buildSrc/src/main/java/com/osacky/flank/gradle/FladleConfig.kt
```diff
@@ -154,22 +159,23 @@ interface FladleConfig {
...
+  val testTargetsAlwaysRun: ListProperty<String>
 
   /**
    * A list of device-path: file-path pairs that indicate the device paths to push files to the device before starting tests, and the paths of files to push.
    * Device paths must be under absolute, whitelisted paths (${EXTERNAL_STORAGE}, or ${ANDROID_DATA}/local/tmp).
    * Source file paths may be in the local filesystem or in Google Cloud Storage (gs://…).
    */
   @get:Input
-  var otherFiles: Map<String, String>
+  @get:Optional
```
runningcode, line 177: why did we add optional here, but not in other properties?

build.gradle
pawelpasterz: Yeah, that was one of my experiments and it was left unintentionally 👍  Thanks

buildSrc/src/main/java/com/osacky/flank/gradle/FladleConfig.kt
pawelpasterz: Nice catch will remove that

sample/build.gradle
```diff
@@ -36,18 +36,18 @@ fladle {
     smartFlankGcsPath = "gs://test-lab-yr9w6qsdvy45q-iurp80dm95h8a/flank/test_app_android.xml"
     configs {
         oranges {
-            useOrchestrator = false
-            testTargets = [
+            useOrchestrator.set(false)
```
runningcode, line 39: this change should not be necesary. If you revert, does it still work? I would like to keep backward compatibility with the groovy dsl.

sample/build.gradle
```diff
@@ -36,18 +36,18 @@ fladle {
...
+            testTargets.set(project.provider { [
```
runningcode, line 40: same here as above and for the rest of the file.

docs/configuration.md
```diff
@@ -76,10 +76,10 @@ fladle {
     smartFlankGcsPath = "gs://tmp_flank/flank/test_app_android.xml"
     configs {
         oranges {
-            useOrchestrator = false
-            testTargets = [
+            useOrchestrator.set(false)
```
runningcode, line 79: thanks for updating the documentation. I believe that this change is not necessary, but we should make this change in the kotlin example in the docs.

buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
```diff
@@ -27,167 +29,92 @@ internal class YamlWriter {
...
+      prop.get().forEach { custom(it) }
     }
   }
 
-  private fun flakyTestAttemptsLine(flakyTestAttempts: Int): String {
-    val label = "num-flaky-test-attempts"
-    return "  $label: $flakyTestAttempts"
-  }
+  private val <T> ListProperty<T>.isPresentAndNotEmpty
+    get() = isPresent && get().isEmpty().not()
```
runningcode, line 114: Is it possible to write it like so?
```suggestion
    get() = isPresent && get().isNotEmpty()
```

buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
```diff
@@ -27,167 +29,92 @@ internal class YamlWriter {
...
 
-  private fun repeatTestsLine(repeatTests: Int): String {
-    val label = "num-test-runs"
-    return "  $label: $repeatTests"
-  }
+  private val <T, K> MapProperty<T, K>.isPresentAndNotEmpty
+    get() = isPresent && get().isEmpty().not()
```
runningcode, line 117: same suggestion here:
```suggestion
    get() = isPresent && get().isNotEmpty()
```

buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
```diff
@@ -1,23 +1,25 @@
...
 internal class YamlWriter {
 
   internal fun createConfigProps(config: FladleConfig, base: FlankGradleExtension): String {
-    if (base.projectId == null) {
+    if (base.projectId.orNull == null) {
       check(base.serviceAccountCredentials.isPresent) { "ServiceAccountCredentials in fladle extension not set. https://github.com/runningcode/fladle#serviceaccountcredentials" }
     }
     check(base.debugApk.isPresent) { "debugApk must be specified" }
-    check(base.instrumentationApk.isPresent xor !base.roboScript.isNullOrBlank()) {
+    check(base.instrumentationApk.isPresent xor base.roboScript.isPresent) {
```
runningcode, line 15: the previous logic checked if the `roboScript` was not null and not blank. Should we check that here as well?

buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
```diff
@@ -27,167 +29,92 @@ internal class YamlWriter {
       if (base.instrumentationApk.isPresent) {
         appendln("  test: ${base.instrumentationApk.get()}")
       }
-      appendln(deviceString)
+      if (config.devices.isPresentAndNotEmpty) appendln(createDeviceString(config.devices.get()))
```
runningcode, line 32: Is the device a required property in flank? What would happen if it were omitted?

I believe the previous logic would alway add it, this logic only add it if it is present. Why the change?

runningcode (changes requested):
Looking good. just some small logic changes I noticed. I haven't finished my review though.

buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
pawelpasterz: Actually not, we can omit devices and `NexusLowRes` will be taken as default. TBH the only required properties (from flank's perspective) is app apk and test apk. I have in my mind to change it in fladle but in separate PR

sample/build.gradle
pawelpasterz: I am afraid it is necessary. Without it, tests are failing. What is more, if you add `fullJunitResult` (which on current master branch is `Property<Boolean>`) to `sample` `build.gradle` like:
```gradle
fladle {
    ...
    configs {
        oranges {
            ...
            flakyTestAttempts = 3
            fullJunitResult = true
        }
        ...
    }
    ...
}
```
it fails with:
```
FAILURE: Build failed with an exception.

* Where:
Build file '/Users/gogo/fladle/sample/build.gradle' line: 44

* What went wrong:
A problem occurred evaluating project ':sample'.
> Cannot cast object 'true' with class 'java.lang.Boolean' to class 'org.gradle.api.provider.Property'
```

docs/configuration.md
pawelpasterz: https://github.com/runningcode/fladle/pull/176#discussion_r493391055

sample/build.gradle
pawelpasterz: https://github.com/runningcode/fladle/pull/176#discussion_r493391055

buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
pawelpasterz: TBH I based my implementation on `instrumentationApk` check, but yeah, for now we should check for path being blank as well.

FYI, this will change with https://github.com/runningcode/fladle/issues/165

buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
pawelpasterz: But I can change implementation to reflect the previous one, and change it with another PR

buildSrc/src/main/java/com/osacky/flank/gradle/FlankGradleExtension.kt
```diff
@@ -55,51 +57,51 @@ open class FlankGradleExtension @Inject constructor(objects: ObjectFactory) : Fl
...
 
-  override var resultsDir: String? = null
+  override val resultsDir: Property<String> = objects.property()
 
   override val additionalTestApks: ListProperty<String> = objects.listProperty()
 
   override val runTimeout: Property<String> = objects.property()
 
-  override val ignoreFailedTests: Property<Boolean> = objects.property()
+  override val ignoreFailedTests: Property<Boolean> = objects.property<Boolean>().convention(false)
```
runningcode, line 80: why was this changed to false here?

buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
runningcode: Makes sense. What version of the API/Android SDK does Flank use? I want to ensure that there are no unexpected changes introduced with this change.

runningcode (changes requested):
Two more questions.
There are a lot of large customers of fladle, and I just want to make sure not to introduce any unexpected changes.

buildSrc/src/main/java/com/osacky/flank/gradle/FlankGradleExtension.kt
pawelpasterz: Oh...again, unintentionally....There were lots of changes and got a bit lost 😅  Will fix that.

pawelpasterz: BTW, this is the default value in flank

buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
pawelpasterz: > Makes sense. What version of the API/Android SDK does Flank use? I want to ensure that there are no unexpected changes introduced with this change.

28

https://github.com/Flank/flank/blob/master/test_runner/src/main/kotlin/ftl/config/FtlConstants.kt#L47-L48

buildSrc/src/main/java/com/osacky/flank/gradle/FlankGradleExtension.kt
pawelpasterz: Yeah, now I remember why I did it like this

https://github.com/runningcode/fladle/blob/7678e819b1b595b113c93e44d22a2b2cca4ca82d/buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt#L44

I just moved the default value

buildSrc/src/main/java/com/osacky/flank/gradle/FladlePluginDelegate.kt
```diff
@@ -119,7 +119,7 @@ class FladlePluginDelegate {
               project.log("Configuring fladle.debugApk from variant ${this@app.name}")
               extension.debugApk.set(this@app.outputFile.absolutePath)
             }
-            if (extension.roboScript == null && !extension.instrumentationApk.isPresent) {
+            if (!extension.roboScript.isPresent && !extension.instrumentationApk.isPresent) {
```
runningcode, line 122: in order to maintain the same logic as before, i think this should be:
```suggestion
            if (extension.roboScript.orNull == null && !extension.instrumentationApk.isPresent) {
```

buildSrc/src/main/java/com/osacky/flank/gradle/FlankGradleExtension.kt
```diff
@@ -20,25 +22,25 @@ open class FlankGradleExtension @Inject constructor(objects: ObjectFactory) : Fl
...
   // Project id is automatically discovered by default. Use this to override the project id.
-  override var projectId: String? = null
+  override val projectId: Property<String> = objects.property()
   override val serviceAccountCredentials: RegularFileProperty = objects.fileProperty()
-  override var useOrchestrator: Boolean = false
-  override var autoGoogleLogin: Boolean = false
-  override var devices: List<Map<String, String>> = listOf(mapOf("model" to "NexusLowRes", "version" to "28"))
+  override val useOrchestrator: Property<Boolean> = objects.property<Boolean>().convention(false)
+  override val autoGoogleLogin: Property<Boolean> = objects.property<Boolean>().convention(false)
+  override val devices: ListProperty<Map<String, String>> = objects.listProperty<Map<String, String>>().convention(listOf(mapOf("model" to "NexusLowRes", "version" to "28")))
```
runningcode, line 29: thanks for keeping this here. perhaps we can remove it in a separate PR.

buildSrc/src/main/java/com/osacky/flank/gradle/FlankGradleExtension.kt
runningcode: Ah that makes sense. I like keeping all the default values here in one place. 👍

buildSrc/src/main/java/com/osacky/flank/gradle/FlankGradleExtension.kt
```diff
@@ -111,41 +113,41 @@ open class FlankGradleExtension @Inject constructor(objects: ObjectFactory) : Fl
...
-      flakyTestAttempts = flakyTestAttempts,
-      directoriesToPull = directoriesToPull,
-      filesToDownload = filesToDownload,
-      environmentVariables = environmentVariables,
-      recordVideo = recordVideo,
-      performanceMetrics = performanceMetrics,
-      resultsBucket = resultsBucket,
-      keepFilePath = keepFilePath,
-      resultsDir = resultsDir,
+      useOrchestrator = objects.property<Boolean>().convention(useOrchestrator),
```
runningcode, line 118: thanks for changing all of these! 👍

buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
```diff
@@ -1,23 +1,25 @@
...
 internal class YamlWriter {
 
   internal fun createConfigProps(config: FladleConfig, base: FlankGradleExtension): String {
-    if (base.projectId == null) {
+    if (base.projectId.orNull == null) {
       check(base.serviceAccountCredentials.isPresent) { "ServiceAccountCredentials in fladle extension not set. https://github.com/runningcode/fladle#serviceaccountcredentials" }
     }
     check(base.debugApk.isPresent) { "debugApk must be specified" }
-    check(base.instrumentationApk.isPresent xor !base.roboScript.isNullOrBlank()) {
+    check(base.instrumentationApk.isPresent xor !base.roboScript.orNull.isNullOrBlank()) {
```
runningcode, line 15: 👍

buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
runningcode, line 32: i think this `isPresentAndNotEmpty` can be removed now.

buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
```diff
@@ -27,167 +29,92 @@ internal class YamlWriter {
...
-        appendln("  - class $it")
-      }
-    }
-
-    appendln("  legacy-junit-result: ${config.legacyJunitResult.get()}")
-    appendln("  full-junit-result: ${config.fullJunitResult.get()}")
-    appendln("  output-style: ${config.outputStyle.get()}")
+    appendProperty(config.testShards, name = "max-test-shards")
+    appendProperty(config.shardTime, name = "shard-time")
+    appendProperty(config.repeatTests, name = "num-test-runs")
```
runningcode, line 43: nice cleanup here

buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
```diff
@@ -27,167 +29,92 @@ internal class YamlWriter {
...
+  }
 
-    if (config.roboDirectives.isNotEmpty()) {
-      appendln("  robo-directives:")
-      config.roboDirectives.forEach {
-        val value = it.getOrElse(2) { "" }.let { stringValue -> if (stringValue.isBlank()) "\"\"" else stringValue }
-        appendln("    ${it[0]}:${it[1]}: $value")
-      }
+  private fun <T> StringBuilder.appendListProperty(
+    prop: ListProperty<T>,
```
runningcode, line 103: In this case, I believe the list property must always be of type string. Can we restrict the API in that way in order to prevent it from being abused or having unintended effects?
```suggestion
  private fun StringBuilder.appendListProperty(
    prop: ListProperty<String>,
```

buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
```diff
@@ -27,167 +29,92 @@ internal class YamlWriter {
...
-    config.networkProfile?.let {
-      appendln("  network-profile: $it")
-    }
+  private fun <T> StringBuilder.appendProperty(prop: Property<T>, name: String) {
+    if (prop.isPresent) appendln("  $name: ${prop.get()}")
+  }
 
-    config.roboScript?.let {
-      appendln("  robo-script: $it")
+  private fun <T, K> StringBuilder.appendMapProperty(
```
runningcode, line 91: In this case, I believe the map property must always be of type string. Can we restrict the API in that way in order to prevent it from being abused or having unintended effects?
```suggestion
  private fun StringBuilder.appendMapProperty(
```

runningcode (commented):
Alright, we're getting pretty close. Sorry its a big PR and its taking me a while to get through. :)
Here's another round of comments.

sample/build.gradle
runningcode: This is bug in Gradle: https://github.com/gradle/gradle/issues/9987 😢

runningcode (approved):

runningcode:
Thanks @pawelpasterz for doing this! i know it was a lot of work!

PR #177, opened 2020-09-25
Support roboTest without roboScript (sanityRobo)

### LINKED ISSUE #165: Support for Robo Tests without Robo script

pawelpasterz:
Fixes #165 

Flank supports roboTest without roboScript (sanityRobo) https://github.com/Flank/flank/pull/1108
> Sanity robo should be ONLY launched when there is NO, robo-script, test, additional-app-test-apks. This reflects gcloud's behavior

# ***TO BE UPDATED***

To run sanity robo test add `-PsanityRobo` to gradle command
```
./gradlew runFlank -PsanityRobo
```

With this property, `fladle` should skip appending config yaml file with `robo-directives`, `robo-script`, `test`, `additional-app-test-apks`
Example:
1) Run `./gradlew -p sample-kotlin writeConfigProps -PsanityRobo` from project root
2) generated config yaml should look (no `test`)
```yml
gcloud:
  app: [path]
  device:
  - model: Pixel2
    version: 26
  - model: Nexus5
    version: 23

  use-orchestrator: true
  auto-google-login: false
  record-video: true
  performance-metrics: true
  timeout: 15m
  environment-variables:
    clearPackageData: true
  test-targets:
  - class com.osacky.flank.gradle.sample.ExampleInstrumentedTest#seeView
  num-flaky-test-attempts: 1

flank:
  smart-flank-gcs-path: gs://test-lab-yr9w6qsdvy45q-iurp80dm95h8a/flank/test_app_android.xml
  keep-file-path: false
  ignore-failed-tests: false
  disable-sharding: false
  smart-flank-disable-upload: false
  legacy-junit-result: false
  full-junit-result: false
  output-style: single
```

### DIFF
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -153,6 +153,7 @@ fun org.gradle.api.publish.maven.MavenPom.configureForFladle(pluginName: String)
 tasks.withType(Test::class.java).configureEach {
   // Test fixtures are stored in here so we should re-run tests if the test projects change.
   inputs.dir("src/test/resources")
+  maxParallelForks = Runtime.getRuntime().availableProcessors()
   testLogging {
     events = setOf(TestLogEvent.SKIPPED, TestLogEvent.FAILED, TestLogEvent.PASSED)
   }
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -153,7 +153,6 @@ fun org.gradle.api.publish.maven.MavenPom.configureForFladle(pluginName: String)
 tasks.withType(Test::class.java).configureEach {
   // Test fixtures are stored in here so we should re-run tests if the test projects change.
   inputs.dir("src/test/resources")
-  maxParallelForks = Runtime.getRuntime().availableProcessors()
   testLogging {
     events = setOf(TestLogEvent.SKIPPED, TestLogEvent.FAILED, TestLogEvent.PASSED)
   }
--- a/buildSrc/src/main/java/com/osacky/flank/gradle/FladleConfigImpl.kt
+++ b/buildSrc/src/main/java/com/osacky/flank/gradle/FladleConfigImpl.kt
@@ -49,7 +49,14 @@ data class FladleConfigImpl(
   override val legacyJunitResult: Property<Boolean>,
   override val fullJunitResult: Property<Boolean>
 ) : FladleConfig {
-  fun sanityRoboRun() {
+  /**
+   * Prepare config to run sanity robo.
+   *
+   * Sets [sanityRobo] property as `true`.
+   *
+   * Cleans [instrumentationApk], [additionalTestApks], [roboDirectives], [roboScript] properties.
+   */
+  fun clearPropertiesForSanityRobo() {
     sanityRobo.set(true)
     additionalTestApks.empty()
     instrumentationApk.set("")
--- a/buildSrc/src/main/java/com/osacky/flank/gradle/FladleConfigImpl.kt
+++ b/buildSrc/src/main/java/com/osacky/flank/gradle/FladleConfigImpl.kt
@@ -9,6 +9,9 @@ data class FladleConfigImpl(
   internal val name: String,
   override val projectId: Property<String>,
   override val serviceAccountCredentials: RegularFileProperty,
+  override val debugApk: Property<String>,
+  override val instrumentationApk: Property<String>,
+  override val sanityRobo: Property<Boolean>,
   override val useOrchestrator: Property<Boolean>,
   override val autoGoogleLogin: Property<Boolean>,
   override val devices: ListProperty<Map<String, String>>,
@@ -45,4 +48,12 @@ data class FladleConfigImpl(
   override val outputStyle: Property<String>,
   override val legacyJunitResult: Property<Boolean>,
   override val fullJunitResult: Property<Boolean>
-) : FladleConfig
+) : FladleConfig {
+  fun sanityRoboRun() {
+    sanityRobo.set(true)
+    additionalTestApks.empty()
+    instrumentationApk.set("")
+    roboDirectives.empty()
+    roboScript.set("")
+  }
+}
--- a/buildSrc/src/main/java/com/osacky/flank/gradle/FladlePluginDelegate.kt
+++ b/buildSrc/src/main/java/com/osacky/flank/gradle/FladlePluginDelegate.kt
@@ -40,6 +40,9 @@ class FladlePluginDelegate {
       // Must be done afterEvaluate otherwise extension values will not be set.
       project.dependencies.add(FLADLE_CONFIG, "${base.flankCoordinates.get()}:${base.flankVersion.get()}")
 
+      checkIfSanityAndValidateConfigs(base)
+      base.configs.forEach(::checkIfSanityAndValidateConfigs)
+
       // Only use automatic apk path detection for 'com.android.application' projects.
       project.pluginManager.withPlugin("com.android.application") {
         if (!base.debugApk.isPresent || !base.instrumentationApk.isPresent) {
--- a/buildSrc/src/main/java/com/osacky/flank/gradle/FladlePluginDelegate.kt
+++ b/buildSrc/src/main/java/com/osacky/flank/gradle/FladlePluginDelegate.kt
@@ -40,6 +40,8 @@ class FladlePluginDelegate {
       // Must be done afterEvaluate otherwise extension values will not be set.
       project.dependencies.add(FLADLE_CONFIG, "${base.flankCoordinates.get()}:${base.flankVersion.get()}")
 
+      if (project.hasProperty("sanityRobo")) base.sanityRobo.set(true)
+
       // Only use automatic apk path detection for 'com.android.application' projects.
       project.pluginManager.withPlugin("com.android.application") {
         if (!base.debugApk.isPresent || !base.instrumentationApk.isPresent) {
--- a/buildSrc/src/main/java/com/osacky/flank/gradle/FladlePluginDelegate.kt
+++ b/buildSrc/src/main/java/com/osacky/flank/gradle/FladlePluginDelegate.kt
@@ -40,8 +40,6 @@ class FladlePluginDelegate {
       // Must be done afterEvaluate otherwise extension values will not be set.
       project.dependencies.add(FLADLE_CONFIG, "${base.flankCoordinates.get()}:${base.flankVersion.get()}")
 
-      if (project.hasProperty("sanityRobo")) base.sanityRobo.set(true)
-
       // Only use automatic apk path detection for 'com.android.application' projects.
       project.pluginManager.withPlugin("com.android.application") {
         if (!base.debugApk.isPresent || !base.instrumentationApk.isPresent) {
--- a/buildSrc/src/main/java/com/osacky/flank/gradle/FlankGradleExtension.kt
+++ b/buildSrc/src/main/java/com/osacky/flank/gradle/FlankGradleExtension.kt
@@ -20,96 +20,58 @@ open class FlankGradleExtension @Inject constructor(objects: ObjectFactory) : Fl
   val flankCoordinates: Property<String> = objects.property(String::class.java).convention("com.github.flank:flank")
 
   @get:Input
-  val sanityRobo: Property<Boolean> = objects.property<Boolean>().convention(false)
-
-  @get:Input
   val flankVersion: Property<String> = objects.property(String::class.java).convention("20.09.3")
-  // Project id is automatically discovered by default. Use this to override the project id.
-  override val projectId: Property<String> = objects.property()
-  override val serviceAccountCredentials: RegularFileProperty = objects.fileProperty()
-  override val useOrchestrator: Property<Boolean> = objects.property<Boolean>().convention(false)
-  override val autoGoogleLogin: Property<Boolean> = objects.property<Boolean>().convention(false)
-  override val devices: ListProperty<Map<String, String>> = objects.listProperty<Map<String, String>>().convention(listOf(mapOf("model" to "NexusLowRes", "version" to "28")))
-
-  // https://cloud.google.com/sdk/gcloud/reference/firebase/test/android/run
-  override val testTargets: ListProperty<String> = objects.listProperty()
-
-  override val testShards: Property<Int> = objects.property()
-  override val shardTime: Property<Int> = objects.property()
-  override val repeatTests: Property<Int> = objects.property()
-
-  // Shard Android tests by time using historical run data. The amount of shards used is set by `testShards`.
-  override val smartFlankGcsPath: Property<String> = objects.property()
-
-  override val resultsHistoryName: Property<String> = objects.property()
-
-  override val flakyTestAttempts: Property<Int> = objects.property<Int>().convention(0)
 
   // Variant to use for configuring output APK.
   @get:Input
   @get:Optional
   val variant: Property<String> = objects.property()
 
-  /**
-   * debugApk and instrumentationApk are [Property<String>] and not [RegularFileProperty] because we support wildcard characters.
-   */
-  @get:Input
-  @get:Optional
-  val debugApk: Property<String> = objects.property()
-  @get:Input
-  @get:Optional
-  val instrumentationApk: Property<String> = objects.property()
+  // Project id is automatically discovered by default. Use this to override the project id.
+  override val projectId: Property<String> = objects.property()
+  override val serviceAccountCredentials: RegularFileProperty = objects.fileProperty()
+  override val useOrchestrator: Property<Boolean> = objects.property<Boolean>().convention(false)
+  override val autoGoogleLogin: Property<Boolean> = objects.property<Boolean>().convention(false)
+  override val devices: ListProperty<Map<String, String>> = objects.listProperty<Map<String, String>>().convention(listOf(mapOf("model" to "NexusLowRes", "version" to "28")))
+  override val sanityRobo: Property<Boolean> = objects.property<Boolean>().convention(false)
 
+  // https://cloud.google.com/sdk/gcloud/reference/firebase/test/android/run
+  override val testTargets: ListProperty<String> = objects.listProperty()
+  override val testShards: Property<Int> = objects.property()
+  override val shardTime: Property<Int> = objects.property()
+  override val repeatTests: Property<Int> = objects.property()
+
+  // Shard Android tests by time using historical run data. The amount of shards used is set by `testShards`.
+  override val smartFlankGcsPath: Property<String> = objects.property()
+  override val resultsHistoryName: Property<String> = objects.property()
+  override val flakyTestAttempts: Property<Int> = objects.property<Int>().convention(0)
+  override val debugApk: Property<String> = objects.property()
+  override val instrumentationApk: Property<String> = objects.property()
   override val directoriesToPull: ListProperty<String> = objects.listProperty()
-
   override val filesToDownload: ListProperty<String> = objects.listProperty()
-
   override val environmentVariables: MapProperty<String, String> = objects.mapProperty()
-
   override val recordVideo: Property<Boolean> = objects.property<Boolean>().convention(true)
-
   override val performanceMetrics: Property<Boolean> = objects.property<Boolean>().convention(true)
-
   override val resultsBucket: Property<String> = objects.property()
-
   override val keepFilePath: Property<Boolean> = objects.property<Boolean>().convention(false)
-
   override val resultsDir: Property<String> = objects.property()
-
   override val additionalTestApks: ListProperty<String> = objects.listProperty()
-
   override val runTimeout: Property<String> = objects.property()
-
   override val ignoreFailedTests: Property<Boolean> = objects.property<Boolean>().convention(false)
-
   override val disableSharding: Property<Boolean> = objects.property<Boolean>().convention(false)
-
   override val smartFlankDisableUpload: Property<Boolean> = objects.property<Boolean>().convention(false)
-
   override val testRunnerClass: Property<String> = objects.property()
-
   override val localResultsDir: Property<String> = objects.property()
-
   override val numUniformShards: Property<Int> = objects.property()
-
   override val clientDetails: MapProperty<String, String> = objects.mapProperty()
-
   override val testTargetsAlwaysRun: ListProperty<String> = objects.listProperty()
-
   override val otherFiles: MapProperty<String, String> = objects.mapProperty()
-
   override val networkProfile: Property<String> = objects.property()
-
   override val roboScript: Property<String> = objects.property()
-
   override val roboDirectives: ListProperty<List<String>> = objects.listProperty()
-
   override val testTimeout: Property<String> = objects.property<String>().convention("15m")
-
   override val outputStyle: Property<String> = objects.property<String>().convention("single")
-
   override val legacyJunitResult: Property<Boolean> = objects.property<Boolean>().convention(false)
-
   override val fullJunitResult: Property<Boolean> = objects.property<Boolean>().convention(false)
 
   @Internal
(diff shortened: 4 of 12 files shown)


pawelpasterz:
Hey @runningcode please, take a look and let me know what do you think

Thanks!

buildSrc/src/main/java/com/osacky/flank/gradle/FladlePluginDelegate.kt
```diff
@@ -40,6 +40,8 @@ class FladlePluginDelegate {
       // Must be done afterEvaluate otherwise extension values will not be set.
       project.dependencies.add(FLADLE_CONFIG, "${base.flankCoordinates.get()}:${base.flankVersion.get()}")
 
+      if (project.hasProperty("sanityRobo")) base.sanityRobo.set(true)
```
runningcode, line 43: Reading project properties in this way is not compatible with the configuration cache. Can you please remove this or [make it compatible with the configuration cache](https://docs.gradle.org/current/userguide/configuration_cache.html#config_cache:requirements:undeclared_gradle_prop_read)?
The order of precedence for this project property also needs to be specified in the documentation.

buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
```diff
@@ -9,22 +9,24 @@ internal class YamlWriter {
       check(base.serviceAccountCredentials.isPresent) { "ServiceAccountCredentials in fladle extension not set. https://github.com/runningcode/fladle#serviceaccountcredentials" }
     }
     check(base.debugApk.isPresent) { "debugApk must be specified" }
-    check(base.instrumentationApk.isPresent xor !base.roboScript.isNullOrBlank()) {
+    check(base.sanityRobo.get() == true || (base.instrumentationApk.isPresent xor !base.roboScript.isNullOrBlank())) {
       """
      Either instrumentationApk file or roboScript file must be specified but not both.
```
runningcode, line 14: This message needs to be updated.

buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
```diff
@@ -9,22 +9,24 @@ internal class YamlWriter {
       check(base.serviceAccountCredentials.isPresent) { "ServiceAccountCredentials in fladle extension not set. https://github.com/runningcode/fladle#serviceaccountcredentials" }
     }
     check(base.debugApk.isPresent) { "debugApk must be specified" }
-    check(base.instrumentationApk.isPresent xor !base.roboScript.isNullOrBlank()) {
+    check(base.sanityRobo.get() == true || (base.instrumentationApk.isPresent xor !base.roboScript.isNullOrBlank())) {
```
runningcode, line 12: If the sanityRobo is set, does the instrumentation APK also need to be present?

What is the expected behavior if both `sanityRobo` is set and `roboScript` is set?

buildSrc/src/test/java/com/osacky/flank/gradle/integration/SanityRoboCheck.kt
```diff
@@ -0,0 +1,328 @@
...
+      |  disable-sharding: false
+      |  smart-flank-disable-upload: false
+      |  legacy-junit-result: false
+      |  full-junit-result: false
+      |  output-style: single
+    """
+    }
+  }
+
+  private infix fun String.readAndCompareWith(block: () -> String) = testProjectRoot
```
runningcode, line 321: can we rename this to `contains`? it is confusing if the method is named `compareWith` but the assertion is performing `contains`

Also to simplify the comparison, you can use the `printYml` task. It will already read the `yml` file that was written to disk so you can just check the gradle build output.

buildSrc/src/test/java/com/osacky/flank/gradle/integration/TestFixtures.kt
```diff
@@ -1,8 +1,27 @@
...
 
+import org.gradle.testkit.runner.GradleRunner
 import org.junit.rules.TemporaryFolder
 import java.io.File
 
 fun TemporaryFolder.setupFixture(fixtureName: String) {
   File(this::class.java.classLoader.getResource(fixtureName)!!.file).copyRecursively(newFile(fixtureName), true)
 }
+
+internal fun makeGradleFile(where: TemporaryFolder, stringProvider: () -> String) = where
```
runningcode, line 11: I think this would be more straightforward if the last argument were simply a string instead of a stringProvider.
```suggestion
internal fun writeBuildDotGradle(where: TemporaryFolder, buildScript: String) = where
```

buildSrc/src/test/java/com/osacky/flank/gradle/integration/TestFixtures.kt
```diff
@@ -1,8 +1,27 @@
...
+  .newFile("build.gradle")
+  .writeText(stringProvider().trimMargin())
+
+internal fun gradleRun(block: GradleRunFladle.() -> Unit) = GradleRunFladle().apply(block).run {
```
runningcode, line 15: Can this be made in to a simple function?
Or maybe I'm missing something. What is the benefit of this syntax?
```suggestion
internal fun gradleRun(arguments: List<String> = emptyList(), projectDir: File? = null))
```

buildSrc/src/test/java/com/osacky/flank/gradle/integration/SanityRoboCheck.kt
```diff
@@ -0,0 +1,328 @@
...
+
+class SanityRoboCheck {
+  @get:Rule
+  var testProjectRoot = TemporaryFolder()
+
+  @Before
+  fun setUp() = testProjectRoot.newFile("flank-gradle-service.json").writeText("{}")
+
+  @Test
+  fun checkSanityRoboRunSimpleCase() {
```
runningcode, line 17: instead of simple case, can we be more specific?
```suggestion
  fun checkSanityRoboRunWithProjectProperty() {
```

runningcode (changes requested):
Hey @pawelpasterz. Thanks very much for doing this. A lot of people have been asking for this feature.
I also really appreciate the thorough tests.

buildSrc/src/test/java/com/osacky/flank/gradle/integration/TestFixtures.kt
pawelpasterz: Sure, no problem, will change that 👍

buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
pawelpasterz: > If the sanityRobo is set, does the instrumentation APK also need to be present?
No, that's why I've added `base.sanityRobo.get() == true ||` which basically skip validation for sanity robo run

> What is the expected behavior if both sanityRobo is set and roboScript is set?
Same as for `instrumentationApk` and `additionalTestApks` -- it is not appended to config yml

buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
pawelpasterz: I'll refactor `base.sanityRobo.get() == true` to be in separated `if` condition.

buildSrc/src/main/java/com/osacky/flank/gradle/FladlePluginDelegate.kt
pawelpasterz: Fladle needs to support 6.2 as a minimum version https://docs.gradle.org/current/javadoc/org/gradle/api/provider/ProviderFactory.html#gradleProperty-java.lang.String-
For now, fladle will rely on extension property

buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
```diff
@@ -9,22 +9,26 @@ internal class YamlWriter {
       check(base.serviceAccountCredentials.isPresent) { "ServiceAccountCredentials in fladle extension not set. https://github.com/runningcode/fladle#serviceaccountcredentials" }
     }
     check(base.debugApk.isPresent) { "debugApk must be specified" }
-    check(base.instrumentationApk.isPresent xor !base.roboScript.isNullOrBlank()) {
-      """
+    if (base.sanityRobo.get() == false) {
+      check(base.instrumentationApk.isPresent xor !base.roboScript.isNullOrBlank()) {
```
runningcode, line 13: On second though. I don't think that allowing `sanityRobo` to override other configuraiton options is a sustainable way to develop this plugin and could lead to confusing results.
Imagine if `roboScript` were to override the `instrumenationApk` property instead of disallowing this configuraiton. We would have a seriously confusing mess of `if` statements.

To start, can we disallow using `sanityRobo` and `instrumentationApk` and / or `roboScript` within the same fladle config?

runningcode: We should keep this behavior for sanityRobo:
> "Sanity robo should be ONLY launched when there is NO, robo-script, test, additional-app-test-apks. This reflects gcloud's behavior"
I copied that from here https://github.com/Flank/flank/pull/1108

pawelpasterz:
As discussed with @runningcode changing a bit approach
1. `sanityRobo` won't be able to set via gradle property
2. Sanity robo run will ONLY be launched when ALL conditions are met:
    a) no `instrumentationApk`, `roboScript`, `roboDirectives`, `additionalTestApks` are set
    b) `sanityRobo` is explicitly set true (this will turn off autodetection feature for apks/scripts
    ```groovy
    fladle {
        sanityRobo = true
    }
    ```
3. If `sanityRobo = true` and at least on of `instrumentationApk`, `roboScript`, `roboDirectives`, `additionalTestApks` is set as well, an error will be thrown

runningcode (commented):
i haven't finished looking but just some comments! looks pretty good. i will make some suggestions on the error messages in my next pass.

buildSrc/src/main/java/com/osacky/flank/gradle/FlankGradleExtension.kt
```diff
@@ -115,6 +80,9 @@ open class FlankGradleExtension @Inject constructor(objects: ObjectFactory) : Fl
       name = it,
       projectId = objects.property<String>().convention(projectId),
       serviceAccountCredentials = objects.fileProperty().convention(serviceAccountCredentials),
+      debugApk = objects.property<String>().convention(debugApk),
+      instrumentationApk = objects.property<String>().convention(instrumentationApk),
+      sanityRobo = objects.property<Boolean>().convention(false),
```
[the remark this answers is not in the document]
pawelpasterz: I think we shouldn't. Assuming we have `sanityRobo = true` in the base config that would propagate this to the rest, additional configs. My idea was, every config has it's own `sanityRobo` value (default false) that is independent of the rest. I am thinking about the case  when user want `runFlank` to be sanity robo test and all the rest (`runFlank[name1]`, `runFlank[name2]`) are 'normal' test runs

Of course, I can change it to behave otherwise.

buildSrc/build.gradle.kts
```diff
@@ -153,6 +153,7 @@ fun org.gradle.api.publish.maven.MavenPom.configureForFladle(pluginName: String)
 tasks.withType(Test::class.java).configureEach {
   // Test fixtures are stored in here so we should re-run tests if the test projects change.
   inputs.dir("src/test/resources")
+  maxParallelForks = Runtime.getRuntime().availableProcessors()
```
runningcode, line 156: can we make this change separately? How does it affect the performance of the tests?

buildSrc/src/main/java/com/osacky/flank/gradle/SanityConfigValidation.kt
```diff
@@ -0,0 +1,45 @@
...
+    instrumentationApk.isNotPresentOrBlank -> throwBaseConfigError("instrumentationApk")
+    additionalTestApks.isNotPresentOrEmpty -> throwBaseConfigError("additionalTestApks")
+    roboDirectives.isNotPresentOrEmpty -> throwBaseConfigError("roboDirectives")
+    roboScript.isNotPresentOrBlank -> throwBaseConfigError("roboScript")
+  }
+}
+
+private fun FladleConfigImpl.checkAndValidateConfig() {
+  if (sanityRobo.getOrElse(false)) when {
+    roboDirectives.isNotPresentOrEmpty -> throwAdditionalConfigError("roboDirectives", name)
```
runningcode, line 25: why is the ordering different for the property validation?
can we make both methods the same and just replace an empty base config name `""` with `base`?

Although `base` could be the name of the configuration. I don't have a better idea though.

buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
```diff
@@ -12,30 +12,34 @@ internal class YamlWriter {
...
+    val additionalProperties = writeAdditionalProperties(config, shouldPrintTestAndRobo)
+    val flankProperties = writeFlankProperties(config, shouldPrintTestAndRobo)
 
     return buildString {
       appendln("gcloud:")
-      appendln("  app: ${base.debugApk.get()}")
-      if (base.instrumentationApk.isPresent) {
-        appendln("  test: ${base.instrumentationApk.get()}")
+      appendln("  app: ${config.debugApk.get()}")
+      // We don't want to print instrumentation apks if sanityRobo == true
```
runningcode, line 32: shouldn't we throw an error here if sanityRobo is true?

buildSrc/src/test/java/com/osacky/flank/gradle/integration/SanityRoboTest.kt
```diff
@@ -0,0 +1,396 @@
...
+
+    val result = failedGradleRun(
+      arguments = listOf("printYml"),
+      projectDir = testProjectRoot.root
+    )
+
+    assertThat(result.output).contains("FAILED")
+    assertThat(result.output).contains(baseConfigMessage("instrumentationApk"))
+
+    val resultSanity = failedGradleRun(
```
runningcode, line 52: oh, shouldn't this succeed? or am I missing something?

buildSrc/src/main/java/com/osacky/flank/gradle/FladleConfigImpl.kt
```diff
@@ -45,4 +48,12 @@ data class FladleConfigImpl(
   override val outputStyle: Property<String>,
   override val legacyJunitResult: Property<Boolean>,
   override val fullJunitResult: Property<Boolean>
-) : FladleConfig
+) : FladleConfig {
+  fun sanityRoboRun() {
```
runningcode, line 52: I like this method. Its a good compromise!
Can we rename it to `clearPropertiesForSanityRobo()`? And add a bit of javadoc explaining which fields are being cleared?

buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
```diff
@@ -57,7 +62,7 @@ internal class YamlWriter {
     appendProperty(config.outputStyle, name = "output-style")
   }
 
-  internal fun writeAdditionalProperties(config: FladleConfig): String = buildString {
+  internal fun writeAdditionalProperties(config: FladleConfig, printRobo: Boolean = true): String = buildString {
```
runningcode, line 65: instead of having a boolean passed in to this function, maybe we can just separate it in to two functions? I think that would make the code more readable.
i have an intrinsic dislike of boolean parameters to functions 😆

buildSrc/src/test/java/com/osacky/flank/gradle/integration/SanityRoboTest.kt
```diff
@@ -0,0 +1,396 @@
...
+      |      roboScript.set("path/to/script.json")
+      |    }
+      |  }
+      |}
+    """
+    )
+
+    val expectedMessage = additionalConfigMessage("roboScript", "sanity")
+
+    val result = failedGradleRun(
```
runningcode, line 184: shouldn't this one succeed? or am I missing something?

buildSrc/src/test/java/com/osacky/flank/gradle/integration/SanityWithAutoConfigureTest.kt
```diff
@@ -0,0 +1,321 @@
...
+          smart-flank-disable-upload: false
+          legacy-junit-result: false
+          full-junit-result: false
+          output-style: single
+      """.trimIndent()
+    )
+  }
+
+  @Test
+  fun `test auto configuration with sanityRobo set (base config)`() {
```
runningcode, line 207: very nice test. thank you!

runningcode (commented):
Very nice. I went through all of it. I had a few questions about the test cases and error messages but otherwise the overall approach looks like what we want!

buildSrc/build.gradle.kts
pawelpasterz: > can we make this change separately?

sure

> How does it affect the performance of the tests?

On my machine, `./gradlew test` took 1m 41s (without parallelism) and 1m 10s (with parallelism)
So it's not something crucial (IMO) but a nice-to-have feature.

buildSrc/src/test/java/com/osacky/flank/gradle/integration/SanityRoboTest.kt
pawelpasterz: My idea was to take the fail-fast approach and throw an error during the configuration phase

buildSrc/src/test/java/com/osacky/flank/gradle/integration/SanityRoboTest.kt
pawelpasterz: My idea was to take the fail-fast approach and throw an error during the configuration phase

buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
pawelpasterz: Sanity run validation if the user sets `instrumentationApk` explicitly and this occurs before `findDebugAndInstrumentationApk` (autodetection ) is invoked (in `FlankPluginDelegate`). The autodetection feature sets app apk and test apk only for `base` config and then those values are used for inner (or would should I name those additional ?) configs.

So actually this is applicable only for case when `base` config is intended to be sanity robo run and the rest 'normal'

Logic could be simplified if we would agree that `base` config couldn't be sanity robo

buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
pawelpasterz: Actually, I can remove this logic entirely since now we have `sanityRobo` in `FladleConfig`

buildSrc/src/main/java/com/osacky/flank/gradle/SanityConfigValidation.kt
pawelpasterz: Sure, nice catch. That was old implementation before I've added some new properties to `FladleConfig`, thanks!

pawelpasterz:
Hey @runningcode , ready for another round of review :)

buildSrc/src/main/java/com/osacky/flank/gradle/FladlePluginDelegate.kt
```diff
@@ -40,6 +40,9 @@ class FladlePluginDelegate {
       // Must be done afterEvaluate otherwise extension values will not be set.
       project.dependencies.add(FLADLE_CONFIG, "${base.flankCoordinates.get()}:${base.flankVersion.get()}")
 
+      checkIfSanityAndValidateConfigs(base)
```
runningcode, line 43: let's move this to be the first line in `createTasksForConfig`
that way the next line can be removed as well.

buildSrc/src/main/java/com/osacky/flank/gradle/FlankGradleExtension.kt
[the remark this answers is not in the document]
runningcode: You make a good point here. I think we want consistency though. All other values are inherited. We shouldn't assume that users will not try to have multiple different `sanityRobo` configs which inherit from the base.
Let's pass the value from the base here as well.

buildSrc/src/main/java/com/osacky/flank/gradle/SanityConfigValidation.kt
```diff
@@ -0,0 +1,32 @@
...
+import org.gradle.api.provider.Property
+
+@Throws(GradleException::class)
+fun checkIfSanityAndValidateConfigs(config: FladleConfig) = when (config) {
+  is FlankGradleExtension -> config.checkAndValidateConfig() { option, _ ->
+    "Incorrect [base] configuration. [$option] can't be used together with sanityRobo."
+  }
+  is FladleConfigImpl -> config.checkAndValidateConfig(config.name) { option, name ->
+    "Incorrect [$name] configuration. [$option] can't be used together with sanityRobo. " +
+      "If you want to launch robo test run without robo script place only clearPropertiesForSanityRobo() into [$name] configuration"
```
runningcode, line 14:
```suggestion
      "To configure sanityRobo, add clearPropertiesForSanityRobo() into [$name] configuration"
```
I'm changing the wording here a bit because a robo test run could also be with the robo directives. In that case, these instructions could be confusing. Let's refer specifically to `sanityRobo` here.

buildSrc/src/main/java/com/osacky/flank/gradle/SanityConfigValidation.kt
```diff
@@ -0,0 +1,32 @@
...
+  }
+  else -> throw GradleException("Unable to check for sanity, check config type")
```
runningcode, line 16: Let's throw an `IllegalStateException` and add the `config`'s class name (or just `toString()` to the message.
This will help with debugging down the line.

buildSrc/src/main/java/com/osacky/flank/gradle/SanityConfigValidation.kt
```diff
@@ -0,0 +1,32 @@
...
+}
+
+private fun FladleConfig.checkAndValidateConfig(name: String = "base", message: (String, String) -> String) {
+  if (sanityRobo.getOrElse(false)) when {
+    roboDirectives.isNotPresentOrEmpty -> throw GradleException(message("roboDirectives", name))
```
runningcode, line 21: I think an `IllegalArgumentException` might be better here.

buildSrc/src/main/java/com/osacky/flank/gradle/SanityConfigValidation.kt
```diff
@@ -0,0 +1,32 @@
...
+  }
+  is FladleConfigImpl -> config.checkAndValidateConfig(config.name) { option, name ->
+    "Incorrect [$name] configuration. [$option] can't be used together with sanityRobo. " +
+      "If you want to launch robo test run without robo script place only clearPropertiesForSanityRobo() into [$name] configuration"
+  }
+  else -> throw GradleException("Unable to check for sanity, check config type")
+}
+
+private fun FladleConfig.checkAndValidateConfig(name: String = "base", message: (String, String) -> String) {
+  if (sanityRobo.getOrElse(false)) when {
```
runningcode, line 20: We should just get the default value since it is already set to `false`. Adding a `getOrElse(false)` here is like adding a second default value.
```suggestion
  if (sanityRobo.get()) when {
```

buildSrc/src/test/java/com/osacky/flank/gradle/integration/SanityRoboTest.kt
runningcode: Ah makes sense!

buildSrc/src/test/java/com/osacky/flank/gradle/integration/SanityRoboTest.kt
runningcode: Ah same thing. 👍

runningcode (commented):
Alright, looking very good. Another round of comments and hopefully we can merge this very soon.

runningcode (approved):

runningcode:
I'm going to merge and fix the rest of the comments on master  so we can release this!

PR #183, opened 2020-10-01
Add support for authenticating with a Google account.

### LINKED ISSUE #182: Allow running fladle with user authentication
Flank allows authenticating users without a service account. Could we have flade do the same?
https://github.com/Flank/flank#authenticate-with-a-google-account

Ideally there would be a task that runs and delegates to the flank call

runningcode:
fixes #182


digitalbuddha (approved):
Thanks for getting this in so quickly

PR #184, opened 2020-10-01
Follow up fixes and refactoring for sanityRobo support.

runningcode:


buildSrc/src/main/java/com/osacky/flank/gradle/FladlePluginDelegate.kt
```diff
@@ -45,9 +45,6 @@ class FladlePluginDelegate {
       // Must be done afterEvaluate otherwise extension values will not be set.
       project.dependencies.add(FLADLE_CONFIG, "${base.flankCoordinates.get()}:${base.flankVersion.get()}")
 
-      checkIfSanityAndValidateConfigs(base)
```
runningcode, line 48: @pawelpasterz there's a subtle bug here that since the validation happens before the debug and instrumentation apk are automatically added to the config both sanityRobo= true and the instrumentation apk could be set and it would pass validation.

pawelpasterz: Hey! Actually, I was aware of it (that is the reason for if conditions in `YamlWriterTask`) but I am glad you found a way to improve it! 👍

PR #185, opened 2020-10-02
YamlWriter follow up fixes.

runningcode:
This avoids skipping printing of the instrumentationApk if it is set.


buildSrc/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
```diff
@@ -12,30 +12,28 @@ internal class YamlWriter {
...
 
-    val shouldPrintTestAndRobo = config.sanityRobo.get().not()
     val additionalProperties = writeAdditionalProperties(config)
     val flankProperties = writeFlankProperties(config)
 
     return buildString {
       appendln("gcloud:")
       appendln("  app: ${config.debugApk.get()}")
-      // We don't want to print instrumentation apks if sanityRobo == true
-      if (shouldPrintTestAndRobo && config.instrumentationApk.isPresent) {
```
runningcode, line 38: @pawelpasterz I removed this conditional logic since it could lead to unintended consequences. Thanks for the comment in the other PR.

PR #187, opened 2020-10-05
Run unit test in parallel

pawelpasterz:
Speed up a bit build process by making unit test parallel
https://guides.gradle.org/performance/#running_tests

### DIFF
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -156,4 +156,5 @@ tasks.withType(Test::class.java).configureEach {
   testLogging {
     events = setOf(TestLogEvent.SKIPPED, TestLogEvent.FAILED, TestLogEvent.PASSED)
   }
+  maxParallelForks = (Runtime.getRuntime().availableProcessors() / 2).takeIf { it > 0 } ?: 1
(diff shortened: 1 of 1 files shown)


runningcode:
Thanks for trying this. Can you share two build scans, one with this enabled, one without?
A lot of times for integration tests, the speed is limited by disk bandwidth and not CPU power so I just want to see some numbers that this improves build speeds.

pawelpasterz:
> Thanks for trying this. Can you share two build scans, one with this enabled, one without?
> A lot of times for integration tests, the speed is limited by disk bandwidth and not CPU power so I just want to see some numbers that this improves build speeds.

Sure!
with parallelism https://scans.gradle.com/s/hw4he4pgilqru
without parallelism https://scans.gradle.com/s/jkpuh57cvrmuk

As I wrote, `a bit` 😄  I just thought it's nice-to-have feature

runningcode:
I appreciate this PR. I think the 2 seconds of difference in test times could simply be caused by other programs running at the same time. I'm not sure it is worth integrating this change for that. If you think otherwise, feel free to provide more data supporting this change.

PR #188, opened 2020-10-05
Remove fladle's default values [proposal]

pawelpasterz:
### Fladle should relay on flank's default values. (proposal)
Both fladle and flank have some default values implemented, most of them are the same. The only differences I found:
|Option|Fladle|Flank|
| ------------- | ------------- |------------- |
| `recordVideo`  | `true` |`false`|
| `outputStyle`  | `single`  |`multi`|
| `performanceMetrics`  | `true`  |`false`|

This change would make a bit easier to implement new options and maintain test code

### DIFF
--- a/buildSrc/src/main/java/com/osacky/flank/gradle/FladleConfig.kt
+++ b/buildSrc/src/main/java/com/osacky/flank/gradle/FladleConfig.kt
@@ -32,16 +32,20 @@ interface FladleConfig {
   val sanityRobo: Property<Boolean>
 
   @get:Input
+  @get:Optional
   val useOrchestrator: Property<Boolean>
 
   @get:Input
+  @get:Optional
   val autoGoogleLogin: Property<Boolean>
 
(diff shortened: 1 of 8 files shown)


runningcode:
I like the idea here. One reason for the default values is to provide stable releases and configuration for fladle customers as flank has a tendency to change default values.
Because I would prefer a stable configuration for users of fladle. I would propose we shelve this for now.

We should definitely keep the `outputStyle` default value though as without it, the console output piped through gradle doesn't work correctly without it.

pawelpasterz:
> I like the idea here. One reason for the default values is to provide stable releases and configuration for fladle customers as flank has a tendency to change default values.
> Because I would prefer a stable configuration for users of fladle. I would propose we shelve this for now.
> 
> We should definitely keep the `outputStyle` default value though as without it, the console output piped through gradle doesn't work correctly without it.

Makes sens, thanks!

PR #189, opened 2020-10-05
Implement validation feature

pawelpasterz:
Sometimes companies want to stick to a specific flank version but keep fladle up to date. There might be a situation when the user wants to configure fladle task with options which are not supported in flank version that is used. Fladle should prevent such cases and should throw an error in configuration phase.

#### What was done?
* Implemented `@SinceFlank` annotation -- all new properties should be annotated with flank version (example:  `@get:SinceFlank("20.09.3")`)
* validation feature -- during the configuration phase all used options are listed and `flankVersion` property is compared with each `SinceFlank` annotation value. When unavailable option usage detected (`[option flank version]` > `flankVersion`) exception is thrown (example: `Option outputStyle is available since flank 20.6.0, which is higher than used 20.5.0`)
* add annotations for all options back to flank version `8.0.0`

### DIFF
--- a/buildSrc/src/main/java/com/osacky/flank/gradle/FladleConfig.kt
+++ b/buildSrc/src/main/java/com/osacky/flank/gradle/FladleConfig.kt
@@ -1,5 +1,6 @@
 package com.osacky.flank.gradle
 
+import com.osacky.flank.gradle.validation.SinceFlank
 import org.gradle.api.file.RegularFileProperty
 import org.gradle.api.provider.ListProperty
 import org.gradle.api.provider.MapProperty
@@ -17,6 +18,9 @@ interface FladleConfig {
   @get:InputFile
   @get:Optional
   val serviceAccountCredentials: RegularFileProperty
+
+  @get:Input
+  val flankVersion: Property<String>
 
   /**
    * debugApk and instrumentationApk are [Property<String>] and not [RegularFileProperty] because we support wildcard characters.
@@ -195,6 +199,7 @@ interface FladleConfig {
    * (default: no network shaping; see available profiles listed by the `flank test network-profiles list` command).
    * This feature only works on physical devices.
    */
+  @get:SinceFlank("21.0.0")
   @get:Input
   @get:Optional
   val networkProfile: Property<String>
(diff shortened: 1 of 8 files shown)


buildSrc/src/main/java/com/osacky/flank/gradle/FladleConfig.kt
```diff
@@ -18,6 +19,9 @@ interface FladleConfig {
   @get:Optional
   val serviceAccountCredentials: RegularFileProperty
 
+  @get:Input
+  val flankVersion: Property<String>
```
runningcode, line 23: So far this looks good.
However this change to add the flankVersion to the FladleConfig might not actually work.

Can you verify in a test that we are able to override this version in different configs and it actually works?

That being said, i think this complicates things and I'm not sure its a feature we would like to have. Is there a way to achive this same functionality without exposing the version in this interface?

pawelpasterz: I have a couple of ideas, thanks for the input!

pawelpasterz:
@runningcode Ready for the second round of review 😄

buildSrc/src/main/java/com/osacky/flank/gradle/validation/ValidateOptions.kt
```diff
@@ -0,0 +1,56 @@
...
+      else -> false
+    }
+  }
+  .mapNotNull { property -> properties[property.name]?.let { property to it } }
+  .forEach { (property, version) ->
+    val configFlankVersion = flank.toVersion()
+    if (version > configFlankVersion) throw IllegalStateException("Option ${property.name} is available since flank $version, which is higher than used $configFlankVersion")
+  }
+
+private data class FlankVersion(
```
runningcode, line 25: can you use `org.gradle.util.VersionNumber` instead?

runningcode (changes requested):
looks great! I like the `SinceFlank` as a good way of documenting the code going forward.
Just need to check if we can use VersionNumber from Gradle instead of rolling our own.

pawelpasterz:
> looks great! I like the `SinceFlank` as a good way of documenting the code going forward.
> Just need to check if we can use VersionNumber from Gradle instead of rolling our own.

`VersionNumber` works great, thanks!

runningcode (approved):
Thanks for doing this

PR #190, opened 2020-10-13
Move responsibility to FlankExecutionTask and declare inputs

CristianGM:
It requires to merge #160 

This PR is still in the way of making UI Tests cacheable (opt-in), declaring inputs helps, even if we still are missing the outputs and the most important part, adding the APKs and testAPKs as inputs too.

### DIFF
--- a/buildSrc/src/main/java/com/osacky/flank/gradle/FlankExecutionTask.kt
+++ b/buildSrc/src/main/java/com/osacky/flank/gradle/FlankExecutionTask.kt
@@ -1,19 +1,72 @@
 package com.osacky.flank.gradle
 
+import org.gradle.api.DefaultTask
 import org.gradle.api.file.ProjectLayout
+import org.gradle.api.file.RegularFile
+import org.gradle.api.file.RegularFileProperty
+import org.gradle.api.model.ObjectFactory
+import org.gradle.api.provider.Property
+import org.gradle.api.tasks.Classpath
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.InputFile
+import org.gradle.api.tasks.InputFiles
+import org.gradle.api.tasks.JavaExec
+import org.gradle.api.tasks.PathSensitive
+import org.gradle.api.tasks.PathSensitivity
+import org.gradle.api.tasks.TaskAction
+import org.gradle.kotlin.dsl.property
+import org.gradle.process.ExecOperations
 import javax.inject.Inject
 
-open class FlankExecutionTask @Inject constructor(projectLayout: ProjectLayout, private val config: FladleConfig) : FlankJavaExec(projectLayout) {
+open class FlankExecutionTask @Inject constructor(
+  objectFactory: ObjectFactory,
+  private val projectLayout: ProjectLayout,
+  private val execOperations: ExecOperations
+) : DefaultTask() {
+  @get:Input
+  val dumpShards: Property<Boolean> = objectFactory.property<Boolean>().convention(false)
+
+  @get:InputFile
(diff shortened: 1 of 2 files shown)


buildSrc/src/main/java/com/osacky/flank/gradle/FlankExecutionTask.kt
```diff
@@ -1,19 +1,72 @@
...
+  val fladleConfig = objectFactory.fileCollection()
 
   init {
-    doFirst {
-      checkFilesExist(config)
+    group = JavaExec.TASK_GROUP
+    description = "Runs instrumentation tests using flank on firebase test lab."
+
+    outputs.upToDateWhen {
+      false
```
CristianGM, line 46: this isn't really require yet because we don't declare any output for this task, it's more a "security" thing to make sure we don't mess it later adding an output before making sure it's absolutely safe, and op-in

buildSrc/src/main/java/com/osacky/flank/gradle/FlankExecutionTask.kt
```diff
@@ -1,19 +1,72 @@
...
 import javax.inject.Inject
 
-open class FlankExecutionTask @Inject constructor(projectLayout: ProjectLayout, private val config: FladleConfig) : FlankJavaExec(projectLayout) {
+open class FlankExecutionTask @Inject constructor(
+  objectFactory: ObjectFactory,
+  private val projectLayout: ProjectLayout,
+  private val execOperations: ExecOperations
+) : DefaultTask() {
+  @get:Input
+  val dumpShards: Property<Boolean> = objectFactory.property<Boolean>().convention(false)
```
runningcode, line 27: This looks great. Declaring the inputs is a good idea.
Can we add all these input properties without injecting `ExecOperations`? What is the benefit to injecting `ExecOperations` here?

CristianGM: That now I don't need to use doFirst declared on the init because we really have a custom task, we are fully in control and maybe eventually we can replace JavaExec with calling Flank methods (using it as a library).

Now the task is doing stuff, which includes JavaExec, before the task was just JavaExec, no control, no option to move behaviour to it, so you had it on the plugin.

CristianGM: Also...didn't we have this same conversation already?
https://github.com/runningcode/fladle/pull/160#issuecomment-674049030

runningcode: Yeah, we did have this conversation before.
Can we separate the two changes so that we can discuss them separately?
I like how you separated the version bump from the pr that uses the version bump. Can we also separate the usage of `ExecOperations` from declaring the inputs so that we can discuss the merits of each one separately.
I'm quite convinced that declaring the inputs is a great idea. I'm not entirely sure about the `ExecOperations` so let's do them separately.

CristianGM: Should I try again to help making tests cacheable? I mean, are you ready to make Fladle require Gradle 6.0?

I was reading the PRs now and I still think FlankExecution is not a JavaExec task, you don't want to expose to your consumers that, they don't need access to all the JavaExec properties, from the client point of view thats an implementation detail that should be hidden.
I think that's simply abusing inheritance when it's not needed (was the only way before Gradle 6.0, but it was released in 2019, I think it's safe to use it)

PR #194, opened 2020-10-21
Write output into specific directory

### LINKED ISSUE #148: Write flank output to a task specific directory
In order to support running multiple simultaneous flank configs, we need to write the flank output to a task specific directory.
This likely involves configuring the command line output to write the files somewhere specific: https://github.com/runningcode/fladle/blob/master/buildSrc/src/main/java/com/osacky/flank/gradle/FladlePluginDelegate.kt#L90

See: https://github.com/runningcode/fladle/issues/141

pawelpasterz:
Fixes #148 

If no `localResultsDir` set fladle will configure tasks (for each configuration) with a working directory specific to each configuration.
This allows delegating the creation of the results directory to the flank. 
```
./build
├── fladle
│   ├── additionaltests
│   │   ├── flank.yml
│   │   └── results
│   │       ├── 2020-10-22_09-05-18.840938_jjYb
│   │       ├── 2020-10-22_09-08-39.804644_GlIW
│   │       └── 2020-10-22_09-20-23.489689_CUlO
│   ├── flank.yml
│   ├── oranges
│   │   ├── flank.yml
│   │   └── results
│   │       └── 2020-10-22_09-23-17.856220_nUGS
│   └── results
│       └── 2020-10-22_09-24-01.277166_Ktpj
```


buildSrc/src/main/java/com/osacky/flank/gradle/results/UniqueName.kt
```diff
@@ -0,0 +1,57 @@
...
+    if (name == "") {
+      it
+    } else {
+      it.dir(name.toLowerCase(Locale.ROOT))
+    }
+  }.get().asFile.absolutePath
+  return configPath + "/results/${uniqueObjectName()}"
+}
+
+private fun uniqueObjectName(): String {
```
runningcode, line 22: can we just use `UUID.randomUUID().toString()`? or is the date important for some reason?

pawelpasterz: Sure we can, I just copy-pasted implementation from flank, just for testing

runningcode: oh i see. in that case do we want to use a similar name? or can we delegate the creation of this to flank somehow?

buildSrc/src/main/java/com/osacky/flank/gradle/FladlePluginDelegate.kt
```diff
@@ -65,6 +66,8 @@ class FladlePluginDelegate {
   private fun TaskContainer.createTasksForConfig(base: FlankGradleExtension, config: FladleConfig, project: Project, name: String) {
     checkIfSanityAndValidateConfigs(config)
     validateOptionsUsed(config = config, flank = base.flankVersion.get())
+    val resultsDir = if (config.localResultsDir.isPresent) null
```
runningcode, line 69: looks pretty good. can we wrap this if statement inside a function?

runningcode (commented):
nice! just some small nits!

buildSrc/src/main/java/com/osacky/flank/gradle/results/UniqueName.kt
pawelpasterz: > or can we delegate the creation of this to flank somehow

That is rather not possible

> in that case do we want to use a similar name?

That was my intention, but it can be whatever you want. I'd like to leave date + time, that's for sure. Rest is up to you

buildSrc/src/main/java/com/osacky/flank/gradle/FladlePluginDelegate.kt
runningcode: actually reading this a second time, should it be like this?
```suggestion
    val resultsDir = if (config.localResultsDir.isPresent) config.localResultsDir.get()
```

buildSrc/src/main/java/com/osacky/flank/gradle/results/UniqueName.kt
runningcode: if there direction we are going towards is to make this task cacheable, we should not be generating unique output directories. ( I know there is some other ongoing work to make this task cacheable)

I think the reason to have the unique output directory is so that the history can be kept? Is that the use case?

pawelpasterz: Ah, I see your concern. I think this implementation is incorrect then. Need to change it

buildSrc/src/main/java/com/osacky/flank/gradle/FladlePluginDelegate.kt
```diff
@@ -65,6 +66,9 @@ class FladlePluginDelegate {
   private fun TaskContainer.createTasksForConfig(base: FlankGradleExtension, config: FladleConfig, project: Project, name: String) {
     checkIfSanityAndValidateConfigs(config)
     validateOptionsUsed(config = config, flank = base.flankVersion.get())
+    val configName = name.toLowerCase()
+    val useDefaultDir = config.localResultsDir.isPresent.not() && configName.isNotBlank()
+    val setUpWorkingDir: JavaExec.() -> Unit = { if (useDefaultDir) workingDir(project.layout.buildDirectory.dir("fladle/$configName")) }
```
runningcode, line 71: woah is this a function declared in line? never seen that before, but since we have control of the `FlankJavaExec` task, can we just define the function in there?

pawelpasterz: Hm...sure I can rewrite it, my intention was, a function that is used to set up working dir is created for each configuration (wraps some params in the scope)

pawelpasterz: > otherwise, i think this makes sense. we'll need some tests for this though since i'm getting quite confused with all the conditional logic :)

That could be hard (at least for me). I don't know how to programmatically verify if working directory is correctly set. Another approach would be to invoke `runFlank[config] -PdumpShards` and check if json is saved into correct dir but that would require CI to be authenticated...

pawelpasterz:
@runningcode I simplified logic a bit and moved set up dir function into `FlankJavaExec`

PR #228, opened 2021-03-11
Added async property

asadsalman:
Added `async` property as supported by Flank and gcloud.

### DIFF
--- a/fladle-plugin/build.gradle.kts
+++ b/fladle-plugin/build.gradle.kts
@@ -1,7 +1,7 @@
 import org.gradle.api.tasks.testing.logging.TestLogEvent
 
 group = "com.osacky.flank.gradle"
-version = "0.16.0-SNAPSHOT"
+version = "0.15.0-SNAPSHOT"
 description = "Easily Scale your Android Instrumentation Tests with Firebase Test Lab with Flank"
 
 repositories {
--- a/fladle-plugin/build.gradle.kts
+++ b/fladle-plugin/build.gradle.kts
@@ -1,7 +1,7 @@
 import org.gradle.api.tasks.testing.logging.TestLogEvent
 
 group = "com.osacky.flank.gradle"
-version = "0.15.0-SNAPSHOT"
+version = "0.16.0-SNAPSHOT"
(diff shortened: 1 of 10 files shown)


fladle-plugin/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
```diff
@@ -83,6 +83,9 @@ internal class YamlWriter {
     appendProperty(config.recordVideo, name = "record-video")
     appendProperty(config.performanceMetrics, name = "performance-metrics")
     appendProperty(config.testTimeout, name = "timeout")
+    if (config.async.isPresent && config.async.get()) {
```
runningcode, line 86: can the existing `appendProperty` method be used here?

runningcode (commented):
Looks great, just one comment.

fladle-plugin/src/main/java/com/osacky/flank/gradle/YamlWriter.kt
asadsalman: I tried using `appendProperty()` directly without checking the new property's presence and value, but it looks like there are tests that assert on the output yml and broke when `async` was added with a default value of `false`.

Instead of changing those tests to handle the async flag, I only add the `async` flag when it's explicitly added as in the `fladle {}` block.

runningcode: interesting. i guess this was written in that way :\

runningcode (approved):
can you add this PR to the changelog?

fladle-plugin/build.gradle.kts
```diff
@@ -1,7 +1,7 @@
 import org.gradle.api.tasks.testing.logging.TestLogEvent
 
 group = "com.osacky.flank.gradle"
-version = "0.15.0-SNAPSHOT"
+version = "0.16.0-SNAPSHOT"
```
runningcode, line 4: 0.15 was never released. let's keep the version number at 0.15

runningcode (approved):

PR #237, opened 2021-03-31
Update Docs for Using Snapshot

AndrewReitz:
Changed `flank_snapshot` to `flank-snapshot`

Docs currently show flank_snapshot but flank publishes it as flank-snapshot`


runningcode (approved):
Thanks!

PR #238, opened 2021-03-31
Allow Flank Snapshot Usage

AndrewReitz:
Validation checks were causing certain configurations to fail because
the snapshot version would become `0.0.0`. This defaults
the snapshot version checks to use the default version of flank.


runningcode (approved):
Very nice. thank you.

runningcode:
Looks like there's a failing test: https://scans.gradle.com/s/7ohen6lcy4rbc/tests/:fladle-plugin:test/com.osacky.flank.gradle.integration.FlankAuthTestTask/testFlankAuth#1

I re-tried it and it still failed. Seems unrelated but could you take a quick look?

PR #251, opened 2021-04-15
Update changelog to include PR #194

rciovati:
The changes introduced in #194 could break some setups. Let's mention it in the changelog.


runningcode (approved):
Thanks @rciovati

PR #255, opened 2021-05-20
Simple change in docs for clarification

tasomaniac:
Although it has a proper link, the project id part looks like it is related to the above step about web authentication. 2 of our team members had this same confusion. That's why I thought this would make it more clear.


runningcode (approved):
Thanks @tasomaniac

PR #256, opened 2021-05-20
Allow not adding certain modules to additionalTestApks.

### LINKED ISSUE #202: Feature Request: Omit modules with fulladle
When using Fulladle, It would be great to have an option to omit modules so the modules `apk` and `testApk` are not added to the fladle config

runningcode:
Modules can now be omitted with
```
fulladleModuleConfig {
  enabled = false
}
```

fixes #202

TODO:
- [ ] Update docs
- [ ] Maybe use `exclude` instead of `enabled`?

### DIFF
--- a/fladle-plugin/src/main/java/com/osacky/flank/gradle/FulladlePlugin.kt
+++ b/fladle-plugin/src/main/java/com/osacky/flank/gradle/FulladlePlugin.kt
@@ -16,6 +16,11 @@ class FulladlePlugin : Plugin<Project> {
 
     val flankGradleExtension = root.extensions.getByType(FlankGradleExtension::class)
 
+    root.subprojects {
+      // Yuck, cross project configuration
+      extensions.create("fulladleModuleConfig", FulladleModuleExtension::class.java)
+    }
+
     val fulladleConfigureTask = root.tasks.register("configureFulladle") {
       doLast {
         root.subprojects {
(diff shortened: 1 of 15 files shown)


fladle-plugin/src/main/java/com/osacky/flank/gradle/FulladlePlugin.kt
```diff
@@ -64,18 +69,21 @@ class FulladlePlugin : Plugin<Project> {
...
-            val library = extensions.getByType<LibraryExtension>()
-            library.testVariants.configureEach {
-              if (file("$projectDir/src/androidTest").exists()) {
-                outputs.configureEach {
-                  flankGradleExtension.additionalTestApks.add(
-                    root.provider {
-                      "- test: $outputFile"
-                    }
-                  )
+            val fulladleModuleExtension = extensions.getByType(FulladleModuleExtension::class.java)
```
asadsalman, line 72: is `extension. getByType()` a good idea? `getByType` will throw exception if a module doesn't have a `FulladleModuleExtension`. I don't know if we should force modules to override values.

runningcode: It'll be added to all subprojects on line 21. Is there a better way to do this?

asadsalman: How about [`findByType()`](https://docs.gradle.org/current/javadoc/org/gradle/api/plugins/ExtensionContainer.html#findByType-java.lang.Class-)? It won't throw an exception if a module doesn't provide the extension, and you won't have to add it to every module.

runningcode: It's a good point, but I would not expect the consumer of the plugin to create the extension when they need it. In the same way that I would not expect the consumer of the Android Gradle Plugin to create the Android extension when they need it. It should be created in each module so that it can be easily configured.
So instead of devs having to write:
`extensions.create("fulladleModuleConfig", FulladleModuleExtension::class.java) {`
thet can instead write:
`fulladleModuleConfig {`

Is that what you mean? or are we discussing different things?

asadsalman: Makes sense, better to add it to every module and let consumers use `fulladleModuleConfig {}`.

runningcode:
Added in the other PR

PR #259, opened 2021-06-11
Removed hardcoded flank snapshot version

asadsalman:
Flank's snapshot release is called [flank_snapshot](https://mvnrepository.com/artifact/com.github.flank/flank/flank_snapshot), instead of hardcoding this inside fladle (which had hardcoded an older version), we can just check if the version ends with the string "snapshot" (case-insensitive).


runningcode (approved):
nice change. tyvm

PR #257, opened 2021-05-21
Added more fields to fulladleModuleConfig

asadsalman:
This PR adds the ability to override certain fields for modules when using Fulladle. 
You can override the following properties for any module:

1. maxTestShards
2. clientDetails
3. environmentVariables
2. debugApk

PR #260, opened 2021-06-29
Added fulladleModuleConfig to app modules

asadsalman:
A previous PR (#257 ) added the ability to override library module's configuration. This PR extends that ability to app modules as well.

This PR also breaks down the configuration of app and library modules in FulladlePlugin into their own separate methods for better readability.

### DIFF
--- a/fladle-plugin/src/main/java/com/osacky/flank/gradle/FulladlePlugin.kt
+++ b/fladle-plugin/src/main/java/com/osacky/flank/gradle/FulladlePlugin.kt
@@ -4,6 +4,7 @@ import com.android.build.gradle.AppExtension
 import com.android.build.gradle.LibraryExtension
 import org.gradle.api.Plugin
 import org.gradle.api.Project
+import org.gradle.api.provider.MapProperty
 import org.gradle.kotlin.dsl.getByType
 
 /**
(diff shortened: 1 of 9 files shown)


fladle-plugin/src/main/java/com/osacky/flank/gradle/FulladlePlugin.kt
```diff
@@ -138,3 +45,127 @@ class FulladlePlugin : Plugin<Project> {
     }
   }
 }
+
+fun configureLibraryModule(project: Project, flankGradleExtension: FlankGradleExtension) = project.run {
+  pluginManager.withPlugin("com.android.library") {
```
runningcode, line 50: is it possible to rely on this logic only and remove the `plugins.findPlugin("com.android.application")` line?

fladle-plugin/src/main/java/com/osacky/flank/gradle/FulladlePlugin.kt
```diff
@@ -138,3 +45,127 @@ class FulladlePlugin : Plugin<Project> {
...
+    val fulladleModuleExtension = extensions.getByType(FulladleModuleExtension::class.java)
+    if (fulladleModuleExtension.enabled.get()) {
```
runningcode, line 52: the other application module code uses `if (!enabled)  return` can we make it consistent in both cases?

fladle-plugin/src/main/java/com/osacky/flank/gradle/FulladlePlugin.kt
```diff
@@ -138,3 +45,127 @@ class FulladlePlugin : Plugin<Project> {
...
+              appendProperty(it, name = "max-test-shards")
+            }
+          }.trimEnd()
+
+          if (strs.isEmpty() || strs[0].isEmpty()) // root module, should not be added as additional test apk
+            return@test
+
+          val clientDetails = mapPropertyToYaml(fulladleModuleExtension.clientDetails, "client-details")
+
+          val environmentVariables = mapPropertyToYaml(fulladleModuleExtension.environmentVariables, "environment-variables")
```
runningcode, line 142: there's a bit of duplicated/copied logic between the two cases, can we extract the common configuration to a single method?

fladle-plugin/src/main/java/com/osacky/flank/gradle/FulladlePlugin.kt
```diff
@@ -138,3 +45,127 @@ class FulladlePlugin : Plugin<Project> {
...
+    }
+  }
+}
+
+fun configureApplicationModule(project: Project, flankGradleExtension: FlankGradleExtension) = project.run {
+  pluginManager.withPlugin("com.android.application") {
+    val fulladleModuleExtension = extensions.getByType(FulladleModuleExtension::class.java)
+    if (!fulladleModuleExtension.enabled.get())
+      return@withPlugin
+    val appExtension = extensions.getByType<AppExtension>()
```
runningcode, line 99: we can probably make the logic more similar in the two cases by using a common base class of `AppExtension` and `LibraryExtension`

fladle-plugin/src/test/resources/android-project2/build.gradle
```diff
@@ -0,0 +1,59 @@
+plugins {
```
runningcode, line 1: is the contents of the file and this directory the same as what is in `android-project`? if so, we can likely just copy the files from there instead of pasting them again here.

runningcode (commented):
Thanks for contributing this feature. It looks good. I just had some code clean up questions.

fladle-plugin/src/test/resources/android-project2/build.gradle
asadsalman: They're not quite the same, `android-project2/build.gradle` has a `fulladleModuleConfig` block. I didn't want to complicate the test by adding logic around copying `android-project` dir and modifying the `build.gradle` in the test set up.

fladle-plugin/src/main/java/com/osacky/flank/gradle/FulladlePlugin.kt
asadsalman: It is possible, `configureLibraryModule` won't do anything when an app module is passed. But I think adding those `findPlugin` checks makes line 28-31 more readable, It makes it obvious that library modules are passed to the `configureLibraryModule` and doesn't assume the reader knows that it doesn't matter if the check doesn't happen.

fladle-plugin/src/main/java/com/osacky/flank/gradle/FulladlePlugin.kt
asadsalman: Sure

fladle-plugin/src/main/java/com/osacky/flank/gradle/FulladlePlugin.kt
asadsalman: Yeah, `mapPropertyToYaml` was an attempt to to extract out duplicate code, since there were two properties using it. I understand more work can be done in deduplication, but there might be diminishing returns.

asadsalman: I extracted out a bit more of the duplicated logic into methods, have a look!

runningcode:
Thanks for the contribution.

PR #262, opened 2021-07-08
Treat app and library modules the same in Fulladle

asadsalman:
This PR merges the methods used to configure app and library modules in Fulladle plugin.

One major problem it solves is handling library modules as root-level app/test APKs (in case there are no app modules, or all app modules are disabled).

Root-level flank app/test apks are both required, and since libraries don't produce app APKs, we first try to set app modules as root level APKs. If none of the app modules are enabled (or if they have no tests), we then try to set libraries as the root-level app. Libraries need to explicitly set `debugApk` in their `fulladleModuleConfig` block for the APK to be consumed.

### DIFF
--- a/docs/multi-module-testing.md
+++ b/docs/multi-module-testing.md
@@ -103,6 +103,15 @@ You may want to exclude a library module from testing when using Fulladle. You c
       enabled.set(false)
     }
     ```
+
+
+### Overriding root-level config
+Flank offers two different ways to specify APKs for testing, one is the *root-level* configuration (under the `gcloud` namespace) and the other is as an `additional-app-test-apks` (under the `flank` namespace).
+
+Fulladle does not provide the ability to control which module ends up as the root-level module or as an additional module. Either one of app modules or library modules can become a root-level module. If a library module ends up as a root-level module, it needs to specify a `debugApk` in its `fladle` or `fulladleModuleConfig` block.
+
+The root-level configuration (e.g. `maxTestShards`) can also be overridden in the `fulladleModuleConfig` block of whatever module gets picked as the root module.
+
 ## Troubleshooting
 Fulladle might still have some rough edges, but we'd love feedback. Please join us in the [Firebase Community Slack](https://firebase.community/) with any feedback you may have.
 You can also file [Fladle Github issues](https://github.com/runningcode/fladle/issues).
--- a/docs/multi-module-testing.md
+++ b/docs/multi-module-testing.md
@@ -106,8 +106,6 @@ You may want to exclude a library module from testing when using Fulladle. You c
 
 
 ### Overriding root-level config
-Flank offers two different ways to specify APKs for testing, one is the *root-level* configuration (under the `gcloud` namespace) and the other is as an `additional-app-test-apks` (under the `flank` namespace).
-
 Fulladle does not provide the ability to control which module ends up as the root-level module or as an additional module. Either one of app modules or library modules can become a root-level module. If a library module ends up as a root-level module, it needs to specify a `debugApk` in its `fladle` or `fulladleModuleConfig` block.
 
 The root-level configuration (e.g. `maxTestShards`) can also be overridden in the `fulladleModuleConfig` block of whatever module gets picked as the root module.
--- a/fladle-plugin/src/main/java/com/osacky/flank/gradle/FulladlePlugin.kt
+++ b/fladle-plugin/src/main/java/com/osacky/flank/gradle/FulladlePlugin.kt
@@ -23,7 +23,7 @@ class FulladlePlugin : Plugin<Project> {
     }
 
     val fulladleConfigureTask = root.tasks.register("configureFulladle") {
-      var allModulesDisabled = true
+      var modulesEnabled = false
       /**
        * we will first configure all app modules
        * then configure all library modules
@@ -41,7 +41,7 @@ class FulladlePlugin : Plugin<Project> {
         root.subprojects {
           if (!hasAndroidTest)
             return@subprojects
-          allModulesDisabled = false
+          modulesEnabled = true
           if (isAndroidAppModule)
             configureModule(this, flankGradleExtension)
         }
@@ -49,12 +49,14 @@ class FulladlePlugin : Plugin<Project> {
         root.subprojects {
           if (!hasAndroidTest)
             return@subprojects
-          allModulesDisabled = false
+          modulesEnabled = true
           if (isAndroidLibraryModule)
             configureModule(this, flankGradleExtension)
         }
-        check(!allModulesDisabled) {
-          "All modules were disabled for testing in fulladleModuleConfig or the enabled modules had no tests"
+
+        check(modulesEnabled) {
+          "All modules were disabled for testing in fulladleModuleConfig or the enabled modules had no tests.\n" +
+                  "Either re-enable modules for testing or add modules with tests."
         }
       }
     }
@@ -222,5 +224,3 @@ fun overrideRootLevelConfigs(flankGradleExtension: FlankGradleExtension, fulladl
     flankGradleExtension.environmentVariables.set(fulladleModuleExtension.environmentVariables.get())
   }
 }
-
-fun setUpRootDebugApk() {}
--- a/fladle-plugin/src/main/java/com/osacky/flank/gradle/FulladlePlugin.kt
+++ b/fladle-plugin/src/main/java/com/osacky/flank/gradle/FulladlePlugin.kt
@@ -56,7 +56,7 @@ class FulladlePlugin : Plugin<Project> {
 
         check(modulesEnabled) {
           "All modules were disabled for testing in fulladleModuleConfig or the enabled modules had no tests.\n" +
-                  "Either re-enable modules for testing or add modules with tests."
+            "Either re-enable modules for testing or add modules with tests."
         }
       }
     }
--- a/fladle-plugin/src/main/java/com/osacky/flank/gradle/FulladlePlugin.kt
+++ b/fladle-plugin/src/main/java/com/osacky/flank/gradle/FulladlePlugin.kt
@@ -214,9 +214,6 @@ fun overrideRootLevelConfigs(flankGradleExtension: FlankGradleExtension, fulladl
   if (fulladleModuleExtension.maxTestShards.orNull != null) {
     flankGradleExtension.maxTestShards.set(fulladleModuleExtension.maxTestShards.get())
   }
-  if (fulladleModuleExtension.maxTestShards.orNull != null) {
-    flankGradleExtension.maxTestShards.set(fulladleModuleExtension.maxTestShards.get())
-  }
   if (fulladleModuleExtension.clientDetails.orNull != null) {
     flankGradleExtension.clientDetails.set(fulladleModuleExtension.clientDetails.get())
   }
--- a/fladle-plugin/src/main/java/com/osacky/flank/gradle/FulladlePlugin.kt
+++ b/fladle-plugin/src/main/java/com/osacky/flank/gradle/FulladlePlugin.kt
@@ -21,31 +21,32 @@ class FulladlePlugin : Plugin<Project> {
       // Yuck, cross project configuration
       extensions.create("fulladleModuleConfig", FulladleModuleExtension::class.java)
     }
-    var allModulesDisabled = true
+
     val fulladleConfigureTask = root.tasks.register("configureFulladle") {
-      doLast {
+      var allModulesDisabled = true
+      doFirst {
         root.subprojects {
-          val fulladleModuleExtension = extensions.findByType(FulladleModuleExtension::class.java) ?: return@subprojects
-          if (!fulladleModuleExtension.enabled.get()) {
+          if (!hasAndroidTest)
             return@subprojects
-          }
           allModulesDisabled = false
+        }
+      }
+
+      doLast {
+        check(!allModulesDisabled) { "All modules were disabled for testing in fulladleModuleConfig or the enabled modules had no tests" }
+        root.subprojects {
           configureModule(this, flankGradleExtension)
         }
       }
     }
 
-    if (allModulesDisabled) {
-      println("All modules were disabled for testing in fulladle")
+    root.tasks.withType(YamlConfigWriterTask::class.java).configureEach {
+      dependsOn(fulladleConfigureTask)
     }
 
     root.afterEvaluate {
       // TODO add other printYml tasks from other configs
       root.tasks.named("printYml").configure {
-        dependsOn(fulladleConfigureTask)
-      }
-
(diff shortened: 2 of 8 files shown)


fladle-plugin/src/main/java/com/osacky/flank/gradle/FulladlePlugin.kt
```diff
@@ -24,115 +23,127 @@ class FulladlePlugin : Plugin<Project> {
     }
 
     val fulladleConfigureTask = root.tasks.register("configureFulladle") {
+      var allModulesDisabled = true
```
runningcode, line 26: I think it is easier to reason about the logic when variables have positive names. Can we rename it like so?
```suggestion
      var modulesConfigured = false
```

fladle-plugin/src/main/java/com/osacky/flank/gradle/FulladlePlugin.kt
```diff
@@ -24,115 +23,127 @@ class FulladlePlugin : Plugin<Project> {
...
+  val testedExtension = extensions.getByType<TestedExtension>()
+  // Only configure the first test variant per module.
+  // Does anyone test more than one variant per module?
+  var addedTestsForModule = false
 
-      outputs.configureEach {
-        val strs = listOf("- test: $outputFile", debugApk, maxTestShards, clientDetails, environmentVariables)
-        writeAdditionalTestApps(strs, flankGradleExtension, rootProject)
-      }
+  // TODO deal with ignored/filtered variants
```
runningcode, line 86: :( i don't like TODOs in code, but it isnt part of this PR. we should explain the limitations instead of having a TODO.

fladle-plugin/src/main/java/com/osacky/flank/gradle/FulladlePlugin.kt
```diff
@@ -166,3 +177,50 @@ fun writeAdditionalTestApps(strs: List<String>, flankGradleExtension: FlankGradl
...
+  get() {
+    val fulladleModuleExtension = extensions.findByType(FulladleModuleExtension::class.java) ?: return false
+    if ((isAndroidLibraryModule || isAndroidAppModule) && !fulladleModuleExtension.enabled.get()) {
+      return false
+    }
+
+    val testedExtension = extensions.getByType<TestedExtension>()
+    var testsFound = true
+    testedExtension.testVariants.configureEach testVariant@{
+      if (!file("$projectDir/src/androidTest").exists()) {
```
runningcode, line 197: this touches a file, so we should make sure this is method is not called during the Gradle configuration phase.

fladle-plugin/src/main/java/com/osacky/flank/gradle/FulladlePlugin.kt
```diff
@@ -166,3 +177,50 @@ fun writeAdditionalTestApps(strs: List<String>, flankGradleExtension: FlankGradl
...
+        println("Ignoring $name test variant in $path: No tests in $projectDir/src/androidTest")
+        testsFound = false
+      }
+      return@testVariant
+    }
+    return testsFound
+  }
+
+fun overrideRootLevelConfigs(flankGradleExtension: FlankGradleExtension, fulladleModuleExtension: FulladleModuleExtension) {
+  // if the root module overrode any value in its fulladleModuleConfig block
```
runningcode, line 207: this seems racy. why do we have to override these local configs from the root? what is the order of precedence?

fladle-plugin/src/main/java/com/osacky/flank/gradle/FulladlePlugin.kt
```diff
@@ -166,3 +177,50 @@ fun writeAdditionalTestApps(strs: List<String>, flankGradleExtension: FlankGradl
...
+  }
+  if (fulladleModuleExtension.clientDetails.orNull != null) {
+    flankGradleExtension.clientDetails.set(fulladleModuleExtension.clientDetails.get())
+  }
+  if (fulladleModuleExtension.environmentVariables.orNull != null) {
+    flankGradleExtension.environmentVariables.set(fulladleModuleExtension.environmentVariables.get())
+  }
+}
+
+fun setUpRootDebugApk() {}
```
runningcode, line 226: is this leftover?

fladle-plugin/src/test/java/com/osacky/flank/gradle/integration/FulladlePluginIntegrationTest.kt
```diff
@@ -241,4 +241,370 @@ class FulladlePluginIntegrationTest {
...
+      fulladleModuleConfig {
+        enabled = true
+        maxTestShards = 7
+      }
+      """.trimIndent()
+    )
+
+    val result = testProjectRoot.gradleRunner()
+      .withArguments(":printYml")
+      .withGradleVersion("6.9")
```
runningcode, line 297: let's remove this version and just test with the build's version.

fladle-plugin/src/test/java/com/osacky/flank/gradle/integration/FulladlePluginIntegrationTest.kt
```diff
@@ -241,4 +241,370 @@ class FulladlePluginIntegrationTest {
...
+    writeBuildGradle(
+      """
+        buildscript {
+            repositories {
+                google()
+                jcenter()
+            }
+
+            dependencies {
+                classpath 'com.android.tools.build:gradle:4.2.1'
```
runningcode, line 271: can we get this version from a local variable to avoid repeating and making upgrades easier?

fladle-plugin/src/test/java/com/osacky/flank/gradle/integration/FulladlePluginIntegrationTest.kt
```diff
@@ -241,4 +241,370 @@ class FulladlePluginIntegrationTest {
       """.trimIndent()
     )
   }
+
+  @Test
+  fun `test root level module overrides with fulladleModuleConfig`() {
```
runningcode, line 246: good test

fladle-plugin/src/main/java/com/osacky/flank/gradle/FulladlePlugin.kt
```diff
@@ -166,3 +177,50 @@ fun writeAdditionalTestApps(strs: List<String>, flankGradleExtension: FlankGradl
...
+      if (!file("$projectDir/src/androidTest").exists()) {
+        println("Ignoring $name test variant in $path: No tests in $projectDir/src/androidTest")
+        testsFound = false
+      }
+      return@testVariant
+    }
+    return testsFound
+  }
+
+fun overrideRootLevelConfigs(flankGradleExtension: FlankGradleExtension, fulladleModuleExtension: FulladleModuleExtension) {
```
runningcode, line 206: this is overriding the local modules values, not the root values. can we rename?

fladle-plugin/src/main/java/com/osacky/flank/gradle/FulladlePlugin.kt
```diff
@@ -24,115 +23,127 @@ class FulladlePlugin : Plugin<Project> {
...
+        if (!flankGradleExtension.instrumentationApk.isPresent) {
+          flankGradleExtension.instrumentationApk.set(rootProject.provider { this@test.outputFile.absolutePath })
+        } else {
+          // Otherwise, let's just add it to the list.
+          strs.add("      test: ${this@test.outputFile}")
+        }
 
-          val clientDetails = mapPropertyToYaml(fulladleModuleExtension.clientDetails, "client-details")
+        if (strs.isEmpty()) {
+          // this is the root module
```
runningcode, line 129: why do we need this? when executing fulladle, can we just always use the root values and when executing from a single module, can we always use the values set there?
otherwise, it should be an error to have both configured.

fladle-plugin/src/main/java/com/osacky/flank/gradle/FulladlePlugin.kt
```diff
@@ -24,115 +23,127 @@ class FulladlePlugin : Plugin<Project> {
...
-          }
-
-          // If the instrumentation apk isn't yet set, let's use this one.
-          if (!flankGradleExtension.instrumentationApk.isPresent) {
-            flankGradleExtension.instrumentationApk.set(rootProject.provider { this@test.outputFile.absolutePath })
-          } else {
-            // Otherwise, let's just add it to the list.
-            strs.add("      test: ${this@test.outputFile}")
+          } else if (project.isAndroidLibraryModule) {
+            // library modules do not produce an app apk and we'll use the one specified in fulladleModuleConfig block
```
runningcode, line 101: this is a good change. thanks

runningcode (commented):
Thanks so much for the changes. Apologies for my delay in reviewing it. I wanted to make sure i sat down and reviewed it thoroughly.

fladle-plugin/src/main/java/com/osacky/flank/gradle/FulladlePlugin.kt
asadsalman: This was already present in the configuration block. I just moved it to another method called from the configuration block. I'm not sure if checking file presence modifies/touches it.

fladle-plugin/src/main/java/com/osacky/flank/gradle/FulladlePlugin.kt
asadsalman: Since we have no way of specifying which module should end up as root, we can instead specify the `fulladleModuleConfig` in every module. Once a module ends up as a root-level module, if its values are overridden, they are used instead of root values.

fladle-plugin/src/main/java/com/osacky/flank/gradle/FulladlePlugin.kt
asadsalman: It is overriding root values. If a module is picked as a root and it specifies values in its `fulladleModuleConfig`, those are used to override the root values.

fladle-plugin/src/main/java/com/osacky/flank/gradle/FulladlePlugin.kt
runningcode: Makes sense. This isn't related then.

fladle-plugin/src/main/java/com/osacky/flank/gradle/FulladlePlugin.kt
```diff
@@ -166,3 +177,50 @@ fun writeAdditionalTestApps(strs: List<String>, flankGradleExtension: FlankGradl
...
+  // then use those values instead
+  if (fulladleModuleExtension.debugApk.orNull != null) {
+    flankGradleExtension.debugApk.set(fulladleModuleExtension.debugApk.get())
+  }
+  if (fulladleModuleExtension.maxTestShards.orNull != null) {
+    flankGradleExtension.maxTestShards.set(fulladleModuleExtension.maxTestShards.get())
+  }
+  if (fulladleModuleExtension.maxTestShards.orNull != null) {
+    flankGradleExtension.maxTestShards.set(fulladleModuleExtension.maxTestShards.get())
+  }
```
runningcode, line 219: It looks like these 3 lines are repeated from the previous 3 lines? Or am I missing something?

fladle-plugin/src/main/java/com/osacky/flank/gradle/FulladlePlugin.kt
runningcode:
```suggestion
  // if the root module has any value set in its fulladleModuleConfig block
```

runningcode: Ok, I think I had a misunderstanding here in my orignal question. i think I understand better what his happening here.
The `FlankGradleExtension` is used at the root and the `FulladleModuleExtension` is per module.
Why do we always want to set the value for the `FulladleModuleExtension` with the value in the `FlankGradleExtension`?
For example, with `maxTestShards` we may have set a high value which is to be used for the main app apk but not the other modules. We would then have to override it in all the other modules?
The `debugApk` makes sense to me to override, but I think I don't fully understand the use case for `maxTestShards`. 

Also, wow would I opt out of overriding the value?

fladle-plugin/src/main/java/com/osacky/flank/gradle/FulladlePlugin.kt
runningcode:
```suggestion
```

fladle-plugin/src/main/java/com/osacky/flank/gradle/FulladlePlugin.kt
runningcode: Makes sense, it is only in the case where it is chosen as the root module that it overrides.

runningcode (approved):
Look good aside from what appears to me like duplicated code. Once that is addressed I will merge.

Please add some documentation either in this PR or in a follow up. Especially what the expected behavior is that you explained to me in slack.
Please also add this to the release notes in this PR or in a follow up.

docs/multi-module-testing.md
```diff
@@ -103,6 +103,15 @@ You may want to exclude a library module from testing when using Fulladle. You c
       enabled.set(false)
     }
     ```
+
+
+### Overriding root-level config
+Flank offers two different ways to specify APKs for testing, one is the *root-level* configuration (under the `gcloud` namespace) and the other is as an `additional-app-test-apks` (under the `flank` namespace).
+
+Fulladle does not provide the ability to control which module ends up as the root-level module or as an additional module. Either one of app modules or library modules can become a root-level module. If a library module ends up as a root-level module, it needs to specify a `debugApk` in its `fladle` or `fulladleModuleConfig` block.
```
runningcode, line 111: You cannot control which module ends up as the root level module but at the same time you need to tell whichever one ends up as the root to have a `debugApk`? How can you do that if you do not know where to add it?

docs/multi-module-testing.md
```diff
@@ -103,6 +103,15 @@ You may want to exclude a library module from testing when using Fulladle. You c
       enabled.set(false)
     }
     ```
+
+
+### Overriding root-level config
+Flank offers two different ways to specify APKs for testing, one is the *root-level* configuration (under the `gcloud` namespace) and the other is as an `additional-app-test-apks` (under the `flank` namespace).
```
runningcode, line 109: I appreciate the insight here but this seems like a implementation detail of flank and not relevant to fladle.

runningcode (commented):
Just two comments on the docs.

docs/multi-module-testing.md
asadsalman: Since we can't predict which module becomes root, we have to add `debugApk` to all modules. If you find that a module isn't picked as root, you can remove its `debugApk` field.

runningcode: We should probably document this. But let's do that in a follow up. I will merge this for now.

PR #265, opened 2021-07-22
Check module is Android before getting TestedExtension

asadsalman:
TestedExtensions only exist inside Android modules so we check Android-ness before fetching the TestedExtension


runningcode (approved):

runningcode:
Is it possible to write a test for this?

runningcode (approved):
Thanks for the test!

PR #266, opened 2021-07-25
Default fulladleModuleConfig maps to null

asadsalman:
Initializing fulladleModuleConfig's maps to null so that we can check for null-ness (instead of emptiness) before using them.


runningcode:
Closed in favor of https://github.com/runningcode/fladle/pull/271

Issue #270, opened 2021-09-10
environmentVariables are not passed to flank.yml


kozaxinan:
We have environmentVairables in fladle block.

```
environmentVariables = [
      "clearPackageData": "true",
      "listener"        : "a.b.c.Foo1" +
               "a.b.c.Foo2"
  ] 
```
environmentVairables are not present in flank.yml and they are not passed to test run on firebase.

runningcode:
Thanks for filing. Do you know if there was a specific version where this stopped working?

kozaxinan:
It works with 0.15.1.
We are using fulladle plugin with flank 21.09.0.
I will test 0.16.0

kozaxinan:
It works on 0.16.0 as well. I couldnt test with 0.16.1.

kozaxinan:
Working versions has a warnings that latest version doesnt have.
```
WARNING: environment-variables set but directories-to-pull is empty, this will result in the coverage file  not downloading to the bucket

```

kozaxinan:
We also don't have any `environment-variables:` in flank yml file even with working scenario. I am not sure how fulladle/fladle/flank passed the `environment-variables:` to firebase/gcloud.

I tried writing test but I couldn't reproduce anything because for us working and non working fladle versions create same yml file. Issue could be in different place not in yml file.

I can provide details if you need.

PR #271, opened 2021-09-15
Do not override root level configs with empty properties

kozaxinan:
Fixes https://github.com/runningcode/fladle/issues/270

Documentation for using different environment variables per test apk 
https://flank.github.io/flank/using_different_environment_variables_in_different_matrices/

Disclamer; I didnt build local snapshot version and I didnt try final result. But I will.

### DIFF
--- a/fladle-plugin/src/test/java/com/osacky/flank/gradle/integration/FulladlePluginIntegrationTest.kt
+++ b/fladle-plugin/src/test/java/com/osacky/flank/gradle/integration/FulladlePluginIntegrationTest.kt
@@ -238,6 +238,10 @@ class FulladlePluginIntegrationTest {
 
         fladle {
           serviceAccountCredentials = project.layout.projectDirectory.file("android-project/flank-gradle-5cf02dc90531.json")
+          environmentVariables = [
+            "clearPackageData": "true", 
+            "listener": "com.osacky.flank.sample.Listener"
+          ]
         }
       """.trimIndent()
     )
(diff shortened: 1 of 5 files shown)


fladle-plugin/src/test/java/com/osacky/flank/gradle/integration/FulladlePluginIntegrationTest.kt
```diff
@@ -290,7 +300,7 @@ class FulladlePluginIntegrationTest {
            test: [0-9a-zA-Z\/_]*/android-project2/build/outputs/apk/androidTest/debug/android-project2-debug-androidTest.apk
            max-test-shards: 5
            environment-variables:
-               "clearPackageData": "true"
+               "clearPackageData": "false"
```
kozaxinan, line 303: A question. " usage in here looks wrong. Shouldnt be it just `clearPagkageData: false`

runningcode: I think you are right. Maybe these integration tests aren't working properly since this other PR has written it the other way: https://github.com/runningcode/fladle/pull/266/files#diff-8c950a3cebeff2c0a82c58117c197724232bb55f3341881c5715c27701437e90R117

runningcode: What is the actual flank.yml file look like?

runningcode:
I think this fixes the same issue as #266 . I like this solution more since it doesn't involve setting the convention to null.
What do you think @asadsalman ?

fladle-plugin/src/test/java/com/osacky/flank/gradle/integration/FulladlePluginIntegrationTest.kt
kozaxinan: Integrations test are bit confusing. It might make sense to split cases and test case by case. I will try to find out actual flank.yml and check removing ""

runningcode (approved):

runningcode:
I'm going to merge this for now. The `"` quotes issue should be fixed separately.

PR #287, opened 2022-01-28
Update robo-directive doc to match the code

changusmc:
The code expects at least 3 elements for subarray https://github.com/runningcode/fladle/blob/master/fladle-plugin/src/main/java/com/osacky/flank/gradle/YamlWriter.kt#L103 and will throw an out of bounds exception if you pass what is shown in the documentation.


runningcode (approved):
Thanks for fixing the doc.

PR #291, opened 2022-03-14
Expose additionalTestApks in YamlConfigWriterTask itself

ZacSweers:
This allows for this input to be manually configured to source its value from a delegate task instead of the FladleConfig properties put in.

To achieve this easily, this also swaps the `@Nested` config handling for Kotlin class delegation instead. This functionally behaves the same but allows for easy overrides of properties like this. Future properties could be added here too.

### DIFF
--- a/fladle-plugin/src/main/java/com/osacky/flank/gradle/YamlConfigWriterTask.kt
+++ b/fladle-plugin/src/main/java/com/osacky/flank/gradle/YamlConfigWriterTask.kt
@@ -36,7 +36,7 @@ open class YamlConfigWriterTask @Inject constructor(
 
   @get:Input
   override val additionalTestApks: ListProperty<String> = objects.listProperty(String::class.java)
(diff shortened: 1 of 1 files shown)


fladle-plugin/src/main/java/com/osacky/flank/gradle/YamlConfigWriterTask.kt
```diff
@@ -16,10 +18,11 @@ import javax.inject.Inject
 @DisableCachingByDefault(because = "The task writes a small file from in memory properties and does not benefit from caching.")
 open class YamlConfigWriterTask @Inject constructor(
   @get:Nested val base: FlankGradleExtension,
-  @get:Nested val config: FladleConfig,
+  config: FladleConfig,
```
runningcode, line 21: I think this means that changes to this configuration object would not be picked up.

Can you clarify what the benefit of configuring it this way is? I'm not understanding the scenario here.

PR #292, opened 2022-03-15
Add variant support to fulladle

marcosholgado:
Currently when using fulladle you cannot specify a variant for a given module which means that you cannot always test the variant that you want to.

This PR adds a new configuration (variant) to fulladle so a variant can be specified for any given module when using fulladle to assure the wanted variant gets tested.


runningcode:
Sorry, didn't get a chance to take a look at this earlier. Will try to take a look this week.

runningcode (approved):
This looks great! Sorry this took so long!

PR #313, opened 2022-11-08
update flank version

adamvduke:


adamvduke:
@runningcode I see various places where versions of flank are defined, but some of them appear to be intentionally older than the latest. Are there other places that need to be updated?

runningcode:
Yeah, some are intentionally older in order to test backward compatibility.

All the versions which are currently set to 22.05.0 should be udpated. The important ones are the one bundles with the plugin which is here: https://github.com/runningcode/fladle/blob/master/gradle/libs.versions.toml#L32 and the one in the docs (the yaml file that you already updated).

adamvduke:
Thanks. I updated the two other places where "22.05.0" was mentioned. I noticed that https://github.com/runningcode/fladle/blob/master/gradle/libs.versions.toml is documented as being generated by a gradle task, but don't see the `refreshVersionsCatalog` task in `$ ./gradlew tasks --all`. Is that file a generate once, and maintain manually going forward?

PR #334, opened 2023-05-23
Update to latest Flank

justintuchek:
Update to latest Flank version for correct test time calculations https://github.com/Flank/flank/pull/2364

Ran test locally, we are already running with this through flankVersion and seems to be working well

### DIFF
--- a/docs/recipes.md
+++ b/docs/recipes.md
@@ -91,11 +91,11 @@ Use Gradle's [dynamic version syntax] to declare a dynamic version.
 
 === "Groovy"
     ``` groovy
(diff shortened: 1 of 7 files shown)


runningcode:
There's a few other place to update here as well can you do a find and replace on that version number in the project?

justintuchek:
Yup! 

I think I got everything

There's a few other trailing versions specified for integration + validation test. If you'd like those too I can push up another commit

Tested locally and seem to be passing, fit within the assertions for everything except the test covering non-existent Flank versions (as expected)

runningcode (approved):

